PR #3, opened 2018-08-16
initial drop

rocketbitz:


README.md
```diff
@@ -0,0 +1,99 @@
...
+$ go get github.com/alpacahq/alpaca-trade-api-go
+```
+
+## Example
+
+In order to call Alpaca's trade API, you need to obtain an API key pair. Replace <key_id> and <secret_key> with what you get from the web console.
+
+### REST example
+
+```
```
umitanuki: Use go syntax highlight

polygon/rest.go
```diff
@@ -0,0 +1,205 @@
...
+
+const (
+	aggURL    = "%v/v1/historic/agg/%v/%v"
+	tradesURL = "%v/v1/historic/trades/%v/%v"
+	quotesURL = "%v/v1/historic/quotes/%v/%v"
+)
+
+var (
+	base string
+	get  = func(u *url.URL) (*http.Response, error) {
```
umitanuki: Weird space here

umitanuki:
As a library, I'm not impressed with limitint it to a singleton model only. There should be a way to create more than one instance of client in a process, and also there should be a way to pass API key as an argument not just env vars.

alpaca/rest.go
```diff
@@ -0,0 +1,472 @@
...
+	if err = unmarshal(resp, asset); err != nil {
+		return nil, err
+	}
+
+	return asset, nil
+}
+
+// ListBarLists returns a list of bar lists corresponding to the provided
+// symbol list, and filtered by the provided parameters.
+func ListBarLists(symbols []string, opts BarListParams) ([]BarList, error) {
```
umitanuki: It doesn't do anything with opts?

polygon/rest.go
rocketbitz: what do you mean weird space? it's go fmt...

rocketbitz:
I don't understand the use-case of multiple clients in the same process? You mean multiple clients using different credentials in the same process? I would argue that users shouldn't be able to do that for their own protection.

umitanuki:
Why do you know that there is *no* use case for that? I don't know who wants to spin up two HTTP server, but certainly net/http package provides a way to do so, next to the package default singleton model.

rocketbitz:
Structure has been updated.

Issue #13, opened 2019-02-12
paper bars endpoint not found


ttt733:
*Initially reported by @ezuhl in the Alpaca-API repository:*

I am using the paper endpoint via your go-lang api.

Request:
limit := 5
param := alpaca.BarListParams{}
param.Timeframe = "day"
//param.StartDt = &start
//param.EndDt = &end
param.Limit = &limit

```
barList ,err  := alpacaClient.ListBarLists([]string{"MMM", "ABT", "ABBV", "ACN"},param)
```
//https://paper-api.alpaca.markets/v1/bars?limit=5&symbols=MMM%2CABT%2CABBV%2CACN

Response:
{"code":40410000,"message":"endpoint not found"}

ttt733:
Resolved with PR #16 .

Issue #4, opened 2018-08-28
Base URL needs to be configurable


umitanuki:
For paper trading and other testing purposes, API base URL should be configurable.

nu11ptr:
Just came here to file this as well, but I see it already is. Base URL should not be an environment variable only, but should be configurable. This isn't just for flexibility, but also to allow multiple instances of the client to be accessing a paper and prod account at the same time.

bobvdvalk:
Something done with this?

ttt733:
Addressed this with PR #14 - sorry for the wait!

PR #14, opened 2019-02-25
Add SetBaseUrl method

ttt733:
Per issues #4 and #13, better in-code configuration of the URL is desired.


alpaca/rest.go
```diff
@@ -62,6 +62,10 @@ type Client struct {
 	credentials *common.APIKey
 }
 
+func SetBaseUrl(baseurl string) {
```
umitanuki: minor: `baseUrl`?

PR #16, opened 2019-02-27
Update handling of response from bars endpoint

ttt733:
This update addresses issues with the handling of the current format of data from Alpaca's `/bars` endpoint.

### DIFF
--- a/alpaca/rest.go
+++ b/alpaca/rest.go
@@ -20,6 +20,7 @@ var (
 	// environment variable set credentials
 	DefaultClient = NewClient(common.Credentials())
 	base          = "https://api.alpaca.markets/"
+	dataUrl		  = "https://data.alpaca.markets/"
 	do            = func(c *Client, req *http.Request) (*http.Response, error) {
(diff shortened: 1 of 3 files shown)


alpaca/rest.go
```diff
@@ -341,7 +342,7 @@ func (c *Client) ListBarLists(symbols []string, opts BarListParams) ([]BarList,
 		vals.Set("limit", strconv.FormatInt(int64(*opts.Limit), 10))
 	}
 
-	u, err := url.Parse(fmt.Sprintf("%v/v1/bars?%v", base, vals.Encode()))
+	u, err := url.Parse(fmt.Sprintf("%sv1/bars/%s?%v", dataUrl, opts.Timeframe, vals.Encode()))
```
umitanuki: How about `path.Join()`? dataUrl is fine but people could forget slash often in env var.

ttt733: None of the other methods use join - will add the / to match the others. If we want to change all of them to use join we can do that in a new PR.

PR #22, opened 2019-04-05
Add streaming entities

gjtorikian:
This PR adds the Stocks Aggregate streaming entity, as defined in https://polygon.io/sockets.


umitanuki:
Thanks for the contribution! Looks good to me.

PR #24, opened 2019-04-13
polygon.io /exchanges endpoint

matteosantama:
I added the ability for a client to request the /exchanges REST endpoint. This lets a client decode the 'Exchange' field in TradeStream messages to a human readable exchange name.


umitanuki:
Looks good! Thanks!

PR #25, opened 2019-05-02
Add GetOrder endpoint

gjtorikian:
This adds the [get order](https://docs.alpaca.markets/api-documentation/web-api/orders/#get-an-order) API to the client.

Although [`get-an-order-by-client-order-id`](https://docs.alpaca.markets/api-documentation/web-api/orders/#get-an-order-by-client-order-id) exists, I was largely copying the existing `cancel-order` endpoint, which uses `orderID string`.


umitanuki:
I'm a bit surprised this was missing. Great catch

PR #27, opened 2019-05-07
Fixed install command and added "encoding/json" import to example whi…

theTulga:
Fixed install command and added "encoding/json" import to the example which was missing


ttt733:
I don't think you should need to `go get` each bit individually. Did `go get github.com/alpacahq/alpaca-trade-api-go` not work for you?

BattulgaTS:
I got this error when I tried installing it.
```
go get github.com/alpacahq/alpaca-trade-api-go
can't load package: package github.com/alpacahq/alpaca-trade-api-go: no Go files in /Users/tulga/go/src/github.com/alpacahq/alpaca-trade-api-go

```
My go version go1.12.4 darwin/amd64
I'm using MacOSX High Sierra
I'm new to GoLang btw lol

umitanuki:
It depends, but now with Go 1.11 onward, `go get` is getting hidden behind scene with go mod.

henry-jackson:
The solution to this issue is using ```go get github.com/alpacahq/alpaca-trade-api-go/...``` (trailing ellipses is the key) to recursively clone the repo as well as all of the dependencies of the `*.go` files within, if you don't already have them on your machine.

Anyone else experiencing a similar issue should refer to [this](https://golang.org/cmd/go/#hdr-Package_lists_and_patterns)

This issue can probably be closed.

uhari03:
Instead of closing this, I say this should instead fix the README.
I had the same issue #64 , but it was sort of dismissed.

umitanuki:
Although you wouldn't need `go get` these days, the instruction just simply didn't work, it I merged the change.

PR #28, opened 2019-05-07
Added missing "FilledQty" field to Order struct.

alejandroreyesj:
Was trying to log partially filled orders and noticed the missing field. Didn't find any guidelines for contributing so I just tried to follow the general coding pattern in the file. Hope this helps!


ttt733:
It does - thanks!

PR #38, opened 2019-07-09
Allow for options to be set

gjtorikian:
For `GetHistoricTrades`, I found the need to provide a starting offset, but the client doesn't let me do so.

I took a rough crack at what that implementation might look like in the form of a new argument. Happy to take suggestions on other implementations, as well as applying this argument to any other method calls that take additional parameters.

### DIFF
--- a/common/options.go
+++ b/common/options.go
@@ -1,0 +1,7 @@
+package common
+
+// Options provides arguments available to various REST endpoints.
+type Options struct {
+	Offset int64
+	Limit  int
+}
--- a/polygon/polygon_test.go
+++ b/polygon/polygon_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


ttt733:
Hello @gjtorikian - rather than using a common parameter object, I would prefer that we have an object with a parameter list for each endpoint. (You can see an example in the ListBarParams struct in entities.go - param structs can go there.) If you'd like to do it that way, feel free to edit this PR, otherwise I'll open a separate PR later tonight and close this one.

gjtorikian:
That makes sense to me, thanks.

gjtorikian:
Hi @ttt733—were you still planning to open a PR for query parameters? If not, no worries, I can do it. I just want to make sure I’m not doing double work.

ttt733:
Thanks for the reminder @gjtorikian - PR opened. Let me know if it's what you were hoping for and I'll merge it in if so.

PR #43, opened 2019-07-16
Add mean reversion example

andrewkim316:
Adds mean-reversion trading algorithm to golang SDK

### DIFF
--- a/examples/mean-reversion.go
+++ b/examples/mean-reversion.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"math"
 	"os"
-	"strconv"
 	"time"
 
 	"github.com/alpacahq/alpaca-trade-api-go/alpaca"
@@ -14,7 +13,6 @@ import (
 
 type alpacaClientContainer struct {
 	client         *alpaca.Client
-	closingPrices  []float64
 	runningAverage float64
 	lastOrder      string
 	timeToClose    float64
@@ -24,8 +22,8 @@ type alpacaClientContainer struct {
 var alpacaClient alpacaClientContainer
 
 func init() {
-	API_KEY := "API_KEY"
-	API_SECRET := "API_SECRET"
+	API_KEY := "YOUR_API_KEY_HERE"
+	API_SECRET := "YOUR_API_SECRET_HERE"
 	BASE_URL := "https://paper-api.alpaca.markets"
 
 	os.Setenv(common.EnvApiKeyID, API_KEY)
@@ -34,7 +32,6 @@ func init() {
 
 	alpacaClient = alpacaClientContainer{
 		alpaca.NewClient(common.Credentials()),
-		[]float64{},
 		0.0,
 		"",
 		0.0,
@@ -57,11 +54,11 @@ func main() {
 		if isOpen {
 			break
 		}
-		time.Sleep(2000 * time.Millisecond)
+		time.Sleep(60000 * time.Millisecond)
 	}
 	fmt.Println("Market Opened.")
 
-	// Get the running average of prices of the last 20 minutes, waiting until we have 20 bars from market open.
+	// Wait until 20 bars of data since market open have been collected.
 	fmt.Println("Waiting for 20 bars...")
 	for {
 		layout := "2006-01-02T15:04:05.000Z"
@@ -78,19 +75,6 @@ func main() {
 	}
 	fmt.Println("We have 20 bars.")
 
-	amtBars := 20
-	bars, err := alpacaClient.client.GetSymbolBars(alpacaClient.stock, alpaca.ListBarParams{Timeframe: "minute", Limit: &amtBars})
-	if err != nil {
-		fmt.Println(err)
-	}
-	for _, bar := range bars {
-		alpacaClient.closingPrices = append(alpacaClient.closingPrices, float64(bar.Close))
-		if len(alpacaClient.closingPrices) > 20 {
-			alpacaClient.closingPrices = alpacaClient.closingPrices[1:]
-		}
-		alpacaClient.runningAverage = ((alpacaClient.runningAverage * float64(len(alpacaClient.closingPrices)-1)) + float64(bar.Close)) / float64(len(alpacaClient.closingPrices))
-	}
-
 	for {
 		alpacaClient.run()
 	}
@@ -101,9 +85,6 @@ func (alp alpacaClientContainer) run() {
 	if alpacaClient.lastOrder != "" {
 		_ = alp.client.CancelOrder(alpacaClient.lastOrder)
 	}
-
-	// Rebalance the portfolio.
-	alp.rebalance()
 
 	// Figure out when the market will close so we can prepare to sell beforehand.
 	clock, _ := alp.client.GetClock()
@@ -123,10 +104,13 @@ func (alp alpacaClientContainer) run() {
 			qty, _ := position.Qty.Float64()
 			qty = math.Abs(qty)
 			alp.submitMarketOrder(int(qty), position.Symbol, orderSide)
-			// Run script again after market close for next trading day.
-			time.Sleep((60000 * 15) * time.Millisecond)
 		}
+		// Run script again after market close for next trading day.
+		fmt.Println("Sleeping until market close (15 minutes).")
+		time.Sleep((60000 * 15) * time.Millisecond)
 	} else {
+		// Rebalance the portfolio.
+		alp.rebalance()
 		time.Sleep(60000 * time.Millisecond)
 	}
 }
@@ -137,7 +121,8 @@ func (alp alpacaClientContainer) awaitMarketOpen() bool {
 	if clock.IsOpen {
 		return true
 	}
-	fmt.Println("spinning")
+	timeToOpen := int(((clock.NextOpen.UnixNano() - clock.Timestamp.UnixNano()) / 1000000000.0) / 60.0)
+	fmt.Printf("%d minutes til next market open.\n", timeToOpen)
 	return false
 }
 
@@ -154,13 +139,14 @@ func (alp alpacaClientContainer) rebalance() {
 	}
 
 	// Get the new updated price and running average.
-	bars, _ := alp.client.GetSymbolBars(alpacaClient.stock, alpaca.ListBarParams{Timeframe: "minute"})
+	amtBars := 20
+	bars, _ := alp.client.GetSymbolBars(alpacaClient.stock, alpaca.ListBarParams{Timeframe: "minute", Limit: &amtBars})
 	currPrice := float64(bars[len(bars)-1].Close)
-	alpacaClient.closingPrices = append(alpacaClient.closingPrices, currPrice)
-	if len(alpacaClient.closingPrices) > 20 {
-		alpacaClient.closingPrices = alpacaClient.closingPrices[1:]
+	alpacaClient.runningAverage = 0.0
+	for _, bar := range bars {
+		alpacaClient.runningAverage += float64(bar.Close)
 	}
-	alpacaClient.runningAverage = ((alpacaClient.runningAverage * float64(len(alpacaClient.closingPrices)-1)) + currPrice) / float64(len(alpacaClient.closingPrices))
+	alpacaClient.runningAverage /= 20.0
 
 	if currPrice > alpacaClient.runningAverage {
 		// Sell our position if the price is above the running average, if any.
@@ -218,14 +204,14 @@ func (alp alpacaClientContainer) submitLimitOrder(qty int, symbol string, price 
 			TimeInForce: "day",
 		})
 		if err == nil {
-			fmt.Println("Limit order of " + "|" + strconv.Itoa(qty) + " " + symbol + " " + side + "|" + " completed.")
+			fmt.Printf("Limit order of | %d %s %s | sent.\n", qty, symbol, side)
 		} else {
-			fmt.Println("Order of " + "|" + strconv.Itoa(qty) + " " + symbol + " " + side + "|" + " did not go through.")
+			fmt.Printf("Order of | %d %s %s | did not go through.\n", qty, symbol, side)
 		}
 		alpacaClient.lastOrder = order.ID
 		return err
 	}
-	fmt.Println("Quantity is <= 0, order order of " + strconv.Itoa(qty) + " " + symbol + " " + side + " not sent")
+	fmt.Printf("Quantity is <= 0, order of | %d %s %s | not sent.\n", qty, symbol, side)
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


examples/mean-reversion.go
```diff
@@ -0,0 +1,273 @@
...
+	timeToClose    float64
+	stock          string
+}
+
+var alpacaClient alpacaClientContainer
+
+func init() {
+	os.Setenv(common.EnvApiKeyID, "API_KEY")
+	os.Setenv(common.EnvApiSecretKey, "API_SECRET")
+	alpaca.SetBaseUrl("https://paper-api.alpaca.markets")
```
umitanuki: Should be given as variable?

examples/mean-reversion.go
```diff
@@ -0,0 +1,273 @@
...
+		alpacaClient.closingPrices = append(alpacaClient.closingPrices, float64(bar.Close))
+		if len(alpacaClient.closingPrices) > 20 {
+			alpacaClient.closingPrices = alpacaClient.closingPrices[1:]
+		}
+		alpacaClient.runningAverage = ((alpacaClient.runningAverage * float64(len(alpacaClient.closingPrices)-1)) + float64(bar.Close)) / float64(len(alpacaClient.closingPrices))
+	}
+
+	for {
+		cRun := make(chan bool)
+		go alpacaClient.run(cRun)
```
umitanuki: Why goroutine?

examples/mean-reversion.go
```diff
@@ -0,0 +1,273 @@
...
+
+// Rebalance our portfolio every minute based off running average data
+func (alp alpacaClientContainer) run(cRun chan bool) {
+	if alpacaClient.lastOrder != "" {
+		_ = alp.client.CancelOrder(alpacaClient.lastOrder)
+	}
+
+	// Rebalance the portfolio
+	cRebalance := make(chan bool)
+	go alp.rebalance(cRebalance)
```
umitanuki: Again I don't see why this is goroutine if you are waiting for the completion...

umitanuki:
Can you add README with the logic behind this algo?

examples/mean-reversion.go
```diff
@@ -0,0 +1,255 @@
...
+	runningAverage float64
+	lastOrder      string
+	timeToClose    float64
+	stock          string
+}
+
+var alpacaClient alpacaClientContainer
+
+func init() {
+	API_KEY := "API_KEY"
```
ttt733: I generally like to have these as like a "replaceme" or "your api key here" value just so it's clearer to the users where their stuff goes.

examples/mean-reversion.go
```diff
@@ -0,0 +1,255 @@
...
+	for {
+		isOpen := alpacaClient.awaitMarketOpen()
+		if isOpen {
+			break
+		}
+		time.Sleep(2000 * time.Millisecond)
+	}
+	fmt.Println("Market Opened.")
+
+	// Get the running average of prices of the last 20 minutes, waiting until we have 20 bars from market open.
```
ttt733: The comment for this block should probably just be `Wait until 20 bars of data since market open have been collected`, and then the part about the running average can be moved down to the next block.

examples/mean-reversion.go
```diff
@@ -0,0 +1,255 @@
...
+			if position.Side == "long" {
+				orderSide = "sell"
+			} else {
+				orderSide = "buy"
+			}
+			qty, _ := position.Qty.Float64()
+			qty = math.Abs(qty)
+			alp.submitMarketOrder(int(qty), position.Symbol, orderSide)
+			// Run script again after market close for next trading day.
+			time.Sleep((60000 * 15) * time.Millisecond)
```
ttt733: This line can be removed

examples/mean-reversion.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+}
+
+// Spin until the market is open.
+func (alp alpacaClientContainer) awaitMarketOpen() bool {
+	clock, _ := alp.client.GetClock()
+	if clock.IsOpen {
+		return true
+	}
+	fmt.Println("spinning")
```
ttt733: Debug line? Seems unnecessary.

examples/mean-reversion.go
```diff
@@ -0,0 +1,255 @@
...
+		bars, _ := alpacaClient.client.GetSymbolBars(alpacaClient.stock, alpaca.ListBarParams{Timeframe: "minute", StartDt: &marketOpen})
+		if len(bars) >= 20 {
+			break
+		} else {
+			time.Sleep(60000 * time.Millisecond)
+		}
+	}
+	fmt.Println("We have 20 bars.")
+
+	amtBars := 20
```
ttt733: Is this block actually necessary? It looks like the running average is already recalculated at the top of `run()`.

examples/mean-reversion.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// Rebalance our portfolio every minute based off running average data.
+func (alp alpacaClientContainer) run() {
+	if alpacaClient.lastOrder != "" {
+		_ = alp.client.CancelOrder(alpacaClient.lastOrder)
+	}
+
+	// Rebalance the portfolio.
+	alp.rebalance()
```
ttt733: Maybe do this in the else block below? Just so at the end of the day it doesn't rebalance and then immediately sell off.

examples/mean-reversion.go
```diff
@@ -0,0 +1,255 @@
...
+	positionVal := 0.0
+	position, err := alp.client.GetPosition(alpacaClient.stock)
+	if err != nil {
+	} else {
+		positionQty = int(position.Qty.IntPart())
+		positionVal, _ = position.MarketValue.Float64()
+	}
+
+	// Get the new updated price and running average.
+	bars, _ := alp.client.GetSymbolBars(alpacaClient.stock, alpaca.ListBarParams{Timeframe: "minute"})
```
ttt733: Should probably put a `Limit: 1` on those params

examples/mean-reversion.go
```diff
@@ -0,0 +1,255 @@
...
+			AccountID:   account.ID,
+			AssetKey:    &symbol,
+			Qty:         decimal.NewFromFloat(float64(qty)),
+			Side:        adjSide,
+			Type:        "limit",
+			LimitPrice:  &limPrice,
+			TimeInForce: "day",
+		})
+		if err == nil {
+			fmt.Println("Limit order of " + "|" + strconv.Itoa(qty) + " " + symbol + " " + side + "|" + " completed.")
```
ttt733: Would rather this be `fmt.Printf("Limit order of | %d %s %s | completed", qty, symbol, side)` or so.

examples/mean-reversion.go
```diff
@@ -0,0 +1,255 @@
...
+		} else {
+			fmt.Println("Order of " + "|" + strconv.Itoa(qty) + " " + symbol + " " + side + "|" + " did not go through.")
+		}
+		alpacaClient.lastOrder = order.ID
```
ttt733: Should the last order ID still be set if there was an error?

examples/mean-reversion.go
```diff
@@ -0,0 +1,255 @@
...
+		return err
+	}
+	fmt.Println("Quantity is <= 0, order order of " + strconv.Itoa(qty) + " " + symbol + " " + side + " not sent")
```
ttt733: "order order" should be fixed.

examples/mean-reversion.go
andrewkim316: Agreed, removed it

examples/mean-reversion.go
andrewkim316: This is done so that the function doesn't print "Market closing soon.  Closing positions" every minute, should I still remove?.  I did move it out of the for loop though, it wasn't supposed to be in there

examples/mean-reversion.go
andrewkim316: Just lets the user know that the algo didn't freeze while it waits for market open, we can remove it if you'd like.

examples/mean-reversion.go
ttt733: As long as it's not in the for loop it's fine.

examples/mean-reversion.go
ttt733: What if we use the calendar API to get the next open time and print like "X minutes to open" (rather than "spinning") based on that value?

andrewkim316: Implemented, looks good

examples/mean-reversion.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// Spin until the market is open.
+func (alp alpacaClientContainer) awaitMarketOpen() bool {
+	clock, _ := alp.client.GetClock()
+	if clock.IsOpen {
+		return true
+	}
+	timeToOpen := int(((clock.NextOpen.UnixNano() - clock.Timestamp.UnixNano()) / 1000000000.0) / 60.0)
+	fmt.Printf("%d minutes til next market open.\n", timeToOpen)
```
umitanuki: till

examples/mean-reversion.go
```diff
@@ -0,0 +1,241 @@
...
+
+// Rebalance our portfolio every minute based off running average data.
+func (alp alpacaClientContainer) run() {
+	if alpacaClient.lastOrder != "" {
+		_ = alp.client.CancelOrder(alpacaClient.lastOrder)
+	}
+
+	// Figure out when the market will close so we can prepare to sell beforehand.
+	clock, _ := alp.client.GetClock()
+	timeToClose := int((clock.NextClose.UnixNano() - clock.Timestamp.UnixNano()) / 1000000)
```
umitanuki: `Unix()` instead of `UnixNano()`?

examples/mean-reversion.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// Get the new updated price and running average.
+	amtBars := 20
+	bars, _ := alp.client.GetSymbolBars(alpacaClient.stock, alpaca.ListBarParams{Timeframe: "minute", Limit: &amtBars})
+	currPrice := float64(bars[len(bars)-1].Close)
+	alpacaClient.runningAverage = 0.0
+	for _, bar := range bars {
+		alpacaClient.runningAverage += float64(bar.Close)
+	}
+	alpacaClient.runningAverage /= 20.0
```
umitanuki: divide by `amtBars`

examples/mean-reversion.go
```diff
@@ -0,0 +1,241 @@
...
+		isOpen := alpacaClient.awaitMarketOpen()
+		if isOpen {
+			break
+		}
+		time.Sleep(60000 * time.Millisecond)
+	}
+	fmt.Println("Market Opened.")
+
+	// Wait until 20 bars of data since market open have been collected.
+	fmt.Println("Waiting for 20 bars...")
```
umitanuki: Would be nice to use variable for 20 and replace all use of constant for this

examples/mean-reversion.go
```diff
@@ -0,0 +1,241 @@
...
+	for {
+		layout := "2006-01-02T15:04:05.000Z"
```
umitanuki: Was it not `time.RFC3339`?

examples/mean-reversion.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/alpacahq/alpaca-trade-api-go/alpaca"
+	"github.com/alpacahq/alpaca-trade-api-go/common"
+	"github.com/shopspring/decimal"
+)
+
+type alpacaClientContainer struct {
+	client         *alpaca.Client
+	runningAverage float64
+	lastOrder      string
+	timeToClose    float64
```
umitanuki: Is this used?

examples/mean-reversion.go
```diff
@@ -0,0 +1,241 @@
...
+// Rebalance our portfolio every minute based off running average data.
+func (alp alpacaClientContainer) run() {
+	if alpacaClient.lastOrder != "" {
+		_ = alp.client.CancelOrder(alpacaClient.lastOrder)
+	}
+
+	// Figure out when the market will close so we can prepare to sell beforehand.
+	clock, _ := alp.client.GetClock()
+	timeToClose := int((clock.NextClose.UnixNano() - clock.Timestamp.UnixNano()) / 1000000)
+	if timeToClose < 60000*15 {
```
umitanuki: `clock.NextClose.Sub(clock.Now) < 15 * time.Minute`

examples/mean-reversion.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+}
+
+// Spin until the market is open.
+func (alp alpacaClientContainer) awaitMarketOpen() bool {
+	clock, _ := alp.client.GetClock()
+	if clock.IsOpen {
+		return true
+	}
+	timeToOpen := int(((clock.NextOpen.UnixNano() - clock.Timestamp.UnixNano()) / 1000000000.0) / 60.0)
```
umitanuki: Use `Sub()`

examples/mean-reversion.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	alpacaClient.runningAverage /= 20.0
+
+	if currPrice > alpacaClient.runningAverage {
+		// Sell our position if the price is above the running average, if any.
+		if positionQty > 0 {
+			fmt.Println("Setting position to zero")
+			alp.submitLimitOrder(positionQty, alpacaClient.stock, currPrice, "sell")
+		} else {
+			fmt.Println("No position in the stock.  No action required.")
```
umitanuki: positionQty < 0 is a short position

examples/mean-reversion.go
```diff
@@ -0,0 +1,241 @@
...
+		// Determine optimal amount of shares based on portfolio and market data.
+		account, _ := alp.client.GetAccount()
+		buyingPower, _ := account.BuyingPower.Float64()
+		positions, _ := alp.client.ListPositions()
+		portfolioVal, _ := account.Cash.Float64()
+		for _, position := range positions {
+			rawVal, _ := position.MarketValue.Float64()
+			portfolioVal += rawVal
+		}
+		portfolioShare := (alpacaClient.runningAverage - currPrice) / currPrice * 200
```
umitanuki: What is 200?

examples/mean-reversion.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/alpacahq/alpaca-trade-api-go/common"
+	"github.com/shopspring/decimal"
+)
+
+type alpacaClientContainer struct {
+	client         *alpaca.Client
+	runningAverage float64
+	lastOrder      string
+	timeToClose    float64
+	stock          string
```
umitanuki: How do you set this stock?

umitanuki: oh hard coded below. Do you want to take command line argument and make this completely runnable code? Currently you need to modify `init()` to even start

examples/mean-reversion.go
andrewkim316: Good idea.  Implemented

examples/mean-reversion.go
andrewkim316: This is what I saw on stack overflow, so idk

examples/mean-reversion.go
andrewkim316: Nope, replaced it with amtBars

examples/mean-reversion.go
andrewkim316: That's directly from the C# algo, it's the strat that they used.

examples/mean-reversion.go
```diff
@@ -0,0 +1,248 @@
...
+		os.Setenv(common.EnvApiKeyID, API_KEY)
+	}
+	if common.Credentials().Secret == "" {
+		os.Setenv(common.EnvApiSecretKey, API_SECRET)
+	}
+	alpaca.SetBaseUrl(BASE_URL)
+
+	// Check if user input a stock, default is AAPL
+	stock := "AAPL"
+	if len(os.Args[1:]) == 1 {
```
umitanuki: Just FYI, there is `flag` package...

PR #44, opened 2019-07-17
Add optional params for GetHistoricTrades

ttt733:
Per request in #38

### DIFF
--- a/polygon/rest.go
+++ b/polygon/rest.go
@@ -140,8 +140,19 @@ func (c *Client) GetHistoricAggregatesV2(
 
 // GetHistoricTrades requests polygon's REST API for historic trades
 // on the provided date .
-func (c *Client) GetHistoricTrades(symbol, date string) (totalTrades *HistoricTrades, err error) {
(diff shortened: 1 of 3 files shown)


polygon/rest.go
```diff
@@ -140,8 +140,19 @@ func (c *Client) GetHistoricAggregatesV2(
...
+	opts *GetHistoricTradesParams) (totalTrades *HistoricTrades, err error) {
+
 	offset := int64(0)
+	limit := int64(10000)
+	if opts != nil {
+		offset = opts.Offset
+		if opts.Limit != 0 {
+			limit = opts.Limit
+		}
+	}
```
gjtorikian: I feel like there should be a `if opts.Offset != 0 {` check here too but other than that looks great!

ttt733: A bit unnecessary IMO - would basically just be saying "if it's 0, don't assign the value because it's already 0" and I'd rather just assign it and move on. (I think we save a few CPU instructions that way, but maybe I've just used Python for too long and have become mildly allergic to braces.)

Issue #48, opened 2019-08-07
Better handling for stream errors?


gjtorikian:
Today I received the following error while reading streams:

```
polygon stream read error (websocket: close 1006 (abnormal closure): unexpected EOF)
```

Although the library does attempt to reconnect with `s.conn = openSocket()`, this does not set any of the channels (as `Subscribe`) does, so nothing actually happens.

Since this is just a printed message, and not, for example, a returned `err` value, I can't act on it. I would prefer to reconnect to my established channels on an error like this. Do you have any suggestions on how this can be done? I am not certain the current code will allow for it.

umitanuki:
I see so the reconnection logic should reset all auth/subscription. Potentially you should also be able to receive error by returned value or by registering custom error handler?

gjtorikian:
> by registering custom error handler?

Yes, exactly! I think that is how the old NATS logic worked. 

To be honest I am also not sure why this error even comes up (two days in a row now).

gjtorikian:
The error is very similar to https://github.com/gorilla/websocket/issues/321.

umitanuki:
It could be due to the client side consumption is not fast enough that the server side buffer exceeds certain threshold and disconnects from their side. not sure if this particular indicates it or not.

saaib:
Hi @umitanuki , this topic of 'buffer exceeds and not been able to read too fast" is like talking about Godzilla, everyone knows about it, but nobody knows if it exists.

Can someone please post real technical information about that such buffer and timeouts I have requested it many times on slack and the information has not been provided.

I've got the same error every day, it is nearly impossible to have an algo to complete from market open to close without this issue. This is showing no business viability for a real production roll-out unless you blindly restart your script because you catch an exception, which is very generic and the incorrect way to handle the situation.

Regards,

siran:
I've have disconnect problem, and stop receiving quote updates. The error does not raise an exception that I can catch from my algo. My solution has been to make my algo resilient to restarts, and then restarting the algo every 5min. It's a desperation based 'solution'. BTW, I'm not working with Go, but with Python. But also receive the disconnects ofter.

ttt733:
[...]

PR #49, opened 2019-08-12
Implement re-subscription on websocket reconnection

ttt733:
Addressing the concerns in issue #48 regarding reconnection not re-subscribing to channels.


polygon/stream.go
```diff
@@ -86,6 +87,15 @@ func (s *Stream) Close() error {
 	return s.conn.Close()
 }
 
+func (s *Stream) reconnect() {
+	s.conn = openSocket()
```
umitanuki: You would need `auth()` too?

ttt733: Turns out that there were a few issues that needed fixing, including that.

PR #52, opened 2019-09-04
Add delayed reconnects on socket open failures

ttt733:


polygon/stream.go
```diff
@@ -238,14 +242,22 @@ func openSocket() *websocket.Conn {
...
+		connectionAttempts++
+		c, _, err := websocket.DefaultDialer.Dial(polygonStreamEndpoint, nil)
+		if err != nil {
+			if connectionAttempts == MaxConnectionAttempts {
+				panic(err)
+			}
+		} else {
+			// consume connection message
+			msg := []PolgyonServerMsg{}
+			if err = c.ReadJSON(&msg); err != nil {
```
gjtorikian: Mentioned this in Slack @ttt733 but this should actually be

```go
if err = c.ReadJSON(&msg); err == nil {
```

in order to continue!

ttt733: Yep - releasing a hotfix for that now.

PR #55, opened 2019-09-28
Add trade ID to streams

gjtorikian:
The [trade stream on Polygon](https://polygon.io/sockets) now provides a trade ID:

image


gjtorikian:
@umitanuki Bump? 🙏

umitanuki:
@gjtorikian Thanks!

PR #56, opened 2019-10-01
Remove excess print

gjtorikian:
As far as I can see, this has no bearing on the code, and instead prints out blank spaces.


umitanuki:
Lol goot catch

PR #65, opened 2020-04-01
Establish a separate Polygon API key

gjtorikian:
A paper Alpaca API key no longer works for Polygon. From Quintin Pike, in Slack's #dev-polygon:

> We have upgraded WebSockets to the latest version which has proven much more stable. In doing that the auth service has been rewritten. Apparently, before you could use paper trading credentials. However, that was unintentional and fixed in the new service.

This poses a bit of a problem. I want to consume Polygon data, but I want to keep using my paper account.

The only solution I can think around this is to have the Polygon API key be a separate env var. This can be a prod Alpaca key, while the Alpaca client can continue to use the paper API keys.

/cc @ttt733 @umitanuki

### DIFF
--- a/common/credentials.go
+++ b/common/credentials.go
@@ -14,20 +14,29 @@ const (
 	EnvApiKeyID     = "APCA_API_KEY_ID"
 	EnvApiSecretKey = "APCA_API_SECRET_KEY"
 	EnvApiOAuth     = "APCA_API_OAUTH"
+	EnvPolygonKeyID = "POLY_API_KEY_ID"
 )
 
 type APIKey struct {
-	ID     string
-	Secret string
-	OAuth  string
+	ID           string
+	Secret       string
+	OAuth        string
+	PolygonKeyID string
 }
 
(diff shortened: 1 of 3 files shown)


gjtorikian:
Bump

117:
@gjtorikian this can already be done, use a NewClient for Polygon and provide only the Alpaca live key

```go
import "github.com/alpacahq/alpaca-trade-api-go/common"
import "github.com/alpacahq/alpaca-trade-api-go/polygon"

func main() {
	polygon.NewClient(&common.APIKey{
		ID: "your-paper-key-id-here",
	})
}
```

117:
streaming however is another story and needs changes

gjtorikian:
> streaming however is another story and needs changes

Yes. Hence the PR.

umitanuki:
So it is still useful to have this parameter?

117:
Yeah it solves the problem.

umitanuki:
OK let's merge it then.

PR #68, opened 2020-05-03
misc updates to outdated methods and fields

117:
see https://alpaca.markets/docs/api-documentation/api-v2/assets/#asset-entity

### DIFF
--- a/alpaca/entities.go
+++ b/alpaca/entities.go
@@ -81,13 +81,16 @@ type Position struct {
 }
 
 type Asset struct {
-	ID       string `json:"id"`
-	Name     string `json:"name"`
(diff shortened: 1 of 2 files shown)


117:
when the base urls get parsed it ends up with a double slash currently

```go
urlString := fmt.Sprintf("%s/%s/orders", base, apiVersion)
// https://api.alpaca.markets//v2/orders
```
in the README.md of this repo you can see that the following is used as an example

```go
// no trailing slash works fine
alpaca.SetBaseUrl("https://paper-api.alpaca.markets")
```

umitanuki:
Nice.

PR #72, opened 2020-05-08
Add ExtendedHours into PlaceOrderRequest

jeefy:
As far as I can tell, this is all that's missing to place extended_hours requests via the Go SDK. If I'm missing something, please let me know. :)


gnvk (approved):
LGTM

PR #75, opened 2020-05-24
fix start and end params for rest endpoint

garlicbulb-puzhuo:
Fix the wrong query param names according to https://alpaca.markets/docs/api-documentation/api-v2/market-data/bars/.


117:
I thought it was just me.. nice find.

dbd:
Should the `ListBarParams` struct also be updated to match?
@shlomikushchi is this something that you could integrate?

PR #76, opened 2020-06-01
Adding oto, oco to OrderClass

shlomikushchi:
oto and oco weren't defined in entities.go


rhodee:
Is there an ETA on landing these OrderClass entities?

PR #83, opened 2020-07-22
Feature - Get Order by client_order_id

### LINKED ISSUE #32: Add support for getting orders by client order ID
This endpoint should be supported in the SDK: https://docs.alpaca.markets/api-documentation/web-api/orders/#get-an-order-by-client-order-id

cartman-dev:
Adding GET endpoint for Order using client_order_id [API Reference](https://alpaca.markets/docs/api-documentation/api-v2/orders/#get-an-order-by-client-order-id)

Fixes #32

### DIFF
--- a/alpaca/alpaca_test.go
+++ b/alpaca/alpaca_test.go
@@ -428,6 +428,32 @@ func (s *AlpacaTestSuite) TestAlpaca() {
 		assert.Nil(s.T(), order)
 	}
 
+	// get order by client_order_id
+	{
+		// successful
+		do = func(c *Client, req *http.Request) (*http.Response, error) {
+			order := Order{
+				ClientOrderID: "some_client_order_id",
+			}
+			return &http.Response{
+				Body: genBody(order),
+			}, nil
+		}
+
+		order, err := GetOrderByClientOrderID("some_client_order_id")
+		assert.Nil(s.T(), err)
+		assert.NotNil(s.T(), order)
+
+		// api failure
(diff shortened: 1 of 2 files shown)


sam3000:
Is that the correct API ?   I looked at https://github.com/alpacahq/alpaca-trade-api-python/blob/master/alpaca_trade_api/rest.py#L310 and then did the following locally and it worked:

// GetOrderByClientID submits a request to get an order by the client order ID.
func (c *Client) GetOrderByClientID(clientOrderID string) (*Order, error) {
        u, err := url.Parse(fmt.Sprintf("%s/%s/orders:by_client_order_id", base, apiVersion))
        if err != nil {
                return nil, err
        }

        q := u.Query()
        q.Set("client_order_id", clientOrderID)
        u.RawQuery = q.Encode()

        resp, err := c.get(u)
        if err != nil {
                return nil, err
        }

        order := &Order{}

        if err = unmarshal(resp, order); err != nil {
                return nil, err
        }

        return order, nil
}

umitanuki:
While I see the convinience doing this, it breaks the current design simplicity for 1:1 mapping to endpoints. Also other SDKs need to do the same. It's more on the API side than SDK, I would say? @shlomikushchi what do you think?

shlomikushchi:
Added in this PR: https://github.com/alpacahq/alpaca-trade-api-go/pull/104 according to the implementation in the python SDK (as @sam3000 suggested)

PR #87, opened 2020-08-18
add support for trailing_stop

shlomikushchi:
adding support for creating orders with trailing stop


shlomikushchi:
added the params

PR #91, opened 2020-08-25
WebSocket unsubscribe

negronjl:
This commit adds UnRegister and Unsubscribe to the socket interfaces for both alpaca and polygon


shlomikushchi:
added in: https://github.com/alpacahq/alpaca-trade-api-go/pull/98

PR #92, opened 2020-08-27
Add martingale example

ZippeyKeys12:
Part of article on converting strategies from Python to Go.
I'll update the README once the article is published


shlomikushchi:
thanks!

PR #93, opened 2020-10-04
Add streamUrl to allow proxying streaming connection

dbd:
The PR adds the ability for users to have a separate `dataURL` for REST based market data queries and use a `streamURL` for stream based market data. Separating them allows for easier proxying. The current implementation also allows for backwards compatibility and if someone only sets the `dataURL` it will also set the `streamUrl`

### DIFF
--- a/alpaca/stream.go
+++ b/alpaca/stream.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"net/url"
+	"os"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


shlomikushchi:
Hi, I think it should be simpler.
no need to introduce another url (we already have the base url and the data url)
another url will be confusing, and also it's not the convention in any other sdk
in other SDK I used this env variable to indicate usage of the proxy: `DATA_PROXY_WS`. we should use it here too
we should just check in `stream.go` - does this env variable defined? use that. not defined? use the default data url.

no need to change rest.go

shlomikushchi:
Hi, I have made this PR:https://github.com/alpacahq/alpaca-trade-api-go/pull/94
and tested your code on top of it.
please verify that it works as expected

dbd:
@shlomikushchi works as expected!

PR #98, opened 2020-10-28
feat(unsubscribe): added unsubscribe functionality to stream

Jrenk:
With this PR I added the functionality to unsubscribe specific streaming channels.

Example:
```
// Subscribe to channel
err = stream.Register("T.*", handler)

// Unsubscribe channel
err = stream.Deregister("T.*")
```

This functionality is already possible in the python version of this package.

PR #103, opened 2020-11-30
Add "after" option to ListOrders.

raphael:
Add ability to unregister from streams.
Fix bug with page size option in GetAccountActivities.

### DIFF
--- a/alpaca/alpaca_test.go
+++ b/alpaca/alpaca_test.go
@@ -346,7 +346,7 @@ func (s *AlpacaTestSuite) TestAlpaca() {
 		until := time.Now()
 		limit := 1
 
-		orders, err := ListOrders(&status, &until, &limit, nil)
+		orders, err := ListOrders(&status, nil, &until, &limit, nil)
 		assert.Nil(s.T(), err)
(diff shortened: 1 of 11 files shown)


sam3000:
The Unsubscribe method is nice but should be in a separate PR.
The ListOrders change breaks the API.  I think it'll need something like ListOrdersAfter() instead.

shlomikushchi:
Hi @raphael 
Thanks a lot for the contribution.
as @sam3000 mentioned, you are adding a lot of code in one PR, which makes it difficult to do this merge.
The common way is to create a separate PR for each subject. Easier to review and merge.
There are good things in here there I want to merge, so if you could split this PR to small PRs we could do the iteration of merging. if not, I will manually copy pieces of code so what ever is convenient for you is fine. thanks.

shlomikushchi:
Also, do note that unsubscribe exists already in this PR: https://github.com/alpacahq/alpaca-trade-api-go/pull/98
so there's no need for that code in a separate PR.

raphael:
That makes sense, sorry for the rushed PR. I'll see if I get time to refactor - closing this for now.

Issue #112, opened 2021-01-29
int32 is insufficient for some volume


hfacey:
Please consider updating the type for volume fields in `entities.go` from int32 to int64. int32 has a max value of 2.1 billion, and as we've seen this week, even daily volume can blow past that

eightseventhreethree:
Opened [PR](https://github.com/alpacahq/alpaca-trade-api-go/pull/115/files).

PR #115, opened 2021-02-16
updating volume entity to return int64 for volume.

eightseventhreethree:
Related to [issue](https://github.com/alpacahq/alpaca-trade-api-go/issues/112). Updated tests.


gnvk (approved):
LGTM

PR #116, opened 2021-02-19
Data API v2

gnvk:


examples/v2stream/v2stream.go
```diff
@@ -0,0 +1,58 @@
+package main
```
cocoztho000, line 1: hello, i am testing this example with your branch of the `alpaca-trade-api-go` package  and receive this error:
```
$ go run main.go 
2021/02/28 20:01:41 failed to open Alpaca data stream: failed to WebSocket dial: expected handshake response status code 101 but got 403
2021/02/28 20:01:42 failed to open Alpaca data stream: failed to WebSocket dial: expected handshake response status code 101 but got 403
2021/02/28 20:01:43 failed to open Alpaca data stream: failed to WebSocket dial: expected handshake response status code 101 but got 403
panic: failed to WebSocket dial: expected handshake response status code 101 but got 403

goroutine 1 [running]:
main.main()
        /main.go:32 +0x245
exit status 2
```

Can you double check this is working as expected?

cocoztho000: Hey ignore me, i checked out the wrong commit from your branch.  Testing looks good 👍

gnvk: Can you please do the review, too?

v2/stream/datav2.go
```diff
@@ -0,0 +1,482 @@
...
+	s.wsWriteMutex.Lock()
+	defer s.wsWriteMutex.Unlock()
+
+	if err := s.conn.Close(websocket.StatusNormalClosure, ""); err != nil {
+		return err
+	}
+
+	if final {
+		s.closed.Store(true)
+	}
```
kalmant, line 189: Would it make sense to move this above `if err := s.conn.Close(websocket.StatusNormalClosure, ""); err != nil {`?

I wonder what happens when the user tries to close the stream and receives an error on L183. Is the current outcome what the user would expect?

kalmant (approved):
LGTM

PR #118, opened 2021-03-05
Fix mean-reversion example

vershun:
Fixes a small bug in the mean-reversion example where we mean to sell our shares instead of buy them.


kalmant (approved):
Looks great, thank you!

PR #121, opened 2021-03-31
ENG-4312: Support Notional Value Orders

ducille:


alpaca/entities.go
```diff
@@ -271,7 +272,8 @@ type PortfolioHistory struct {
 type PlaceOrderRequest struct {
 	AccountID     string           `json:"-"`
 	AssetKey      *string          `json:"symbol"`
-	Qty           decimal.Decimal  `json:"qty"`
+	Qty           *decimal.Decimal `json:"qty,omitempty"`
```
ducille, line 275: breaking change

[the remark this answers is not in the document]
gnvk: This is unacceptable in my opinion. See https://golang.org/doc/modules/version-numbers We cannot break backward compatibility in the same minor version (unless it's v0, but we're on v1.8.1).

I see two solutions:
1. Release a new major version: v2. @kalmant is working on a new version of the v2 stream which might also be incompatible. Also, we could remove all polygon leftovers. Plus who knows what. If we decide to go this way we should a) really think through what incompatible changes we want and b) work on a feature branch for now.
2. Be backward compatible. The easiest way I see to do this is to leave the public `PlaceOrderRequest.Qty` non-pointer (`Notional` is fine, adding a new field is backward compatible) and do any necessary conversion (e.g. 0 -> nil) in `client.PlaceOrder()`.

alpaca/alpaca_test.go
```diff
@@ -390,6 +392,27 @@ func (s *AlpacaTestSuite) TestAlpaca() {
...
+		req = PlaceOrderRequest{
+			AccountID:   "some_id",
+			Notional:    &notional,
+			Side:        Buy,
+			TimeInForce: GTC,
+			Type:        Limit,
+		}
+
+		order, err = PlaceOrder(req)
+		assert.Nil(s.T(), err)
```
gnvk, line 411: [minor] please use `assert.NoError` instead

alpaca/entities.go
```diff
@@ -54,7 +54,8 @@ type Order struct {
 	Symbol         string           `json:"symbol"`
 	Exchange       string           `json:"exchange"`
 	Class          string           `json:"asset_class"`
-	Qty            decimal.Decimal  `json:"qty"`
+	Qty            *decimal.Decimal `json:"qty"`
```
gnvk, line 57: Same here.

gnvk (changes requested):
This is backward incompatible in this form.

alpaca/entities.go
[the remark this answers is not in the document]
ducille: Thanks for the feedback, @gnvk!

@kalmant, how far along are you on the v2 stream work? If you're nearly finished and the new version is, in fact, incompatible, we could go with the first solution @gnvk proposed (i.e., release a new major version)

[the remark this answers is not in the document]
kalmant: It is currently being reviewed. It can be added as backward compatible or incompatible as well.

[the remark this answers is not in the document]
gnvk: #123 was only merged for a v2 feature branch. I would recommend to land this on the v1 (now master) in a backward compatible manner if possible.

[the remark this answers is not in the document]
ducille: Thanks @gnvk. I took a first stab at a solution in 4ec00ce. It seems inefficient to me (i.e., marshal -> unmarshal -> marshal), but the alternative (below) seems too prone to breakage:

```go
func (req PlaceOrderRequest) MarshalJSON() ([]byte, error) {
    qty := req.Qty
    notional := req.Notional

    return json.Marshal(struct{
		AccountID     string           `json:"-"`
		AssetKey      *string          `json:"symbol"`
		Qty           *decimal.Decimal `json:"qty,omitempty"`
		Notional      *decimal.Decimal `json:"notional,omitempty"`
		Side          Side             `json:"side"`
		Type          OrderType        `json:"type"`
		TimeInForce   TimeInForce      `json:"time_in_force"`
		LimitPrice    *decimal.Decimal `json:"limit_price"`
		ExtendedHours bool             `json:"extended_hours"`
		StopPrice     *decimal.Decimal `json:"stop_price"`
		ClientOrderID string           `json:"client_order_id"`
		OrderClass    OrderClass       `json:"order_class"`
		TakeProfit    *TakeProfit      `json:"take_profit"`
		StopLoss      *StopLoss        `json:"stop_loss"`
		TrailPrice    *decimal.Decimal `json:"trail_price"`
		TrailPercent  *decimal.Decimal `json:"trail_percent"`
	}{
		AccountID:     req.AccountID,       
		AssetKey:      req.AssetKey,
		Qty:           &qty,
		Notional:      &notional,
		Side:          req.Side,
		Type:          req.Type,
		TimeInForce:   req.TimeInForce,
		LimitPrice:    req.LimitPrice,
		ExtendedHours: req.ExtendedHours,
		StopPrice:     req.StopPrice,
		ClientOrderID: req.ClientOrderID,
		OrderClass:    req.OrderClass,
		TakeProfit:    req.TakeProfit,
		StopLoss:      req.StopLoss,
		TrailPrice:    req.TrailPrice,
		TrailPercent:  req.TrailPercent
    })
}
```

Appreciate any feedback/suggestions you may have

alpaca/alpaca_test.go
```diff
@@ -372,24 +372,45 @@ func (s *AlpacaTestSuite) TestAlpaca() {
...
 			TimeInForce: GTC,
 			Type:        Limit,
 		}
 
 		order, err := PlaceOrder(req)
-		assert.Nil(s.T(), err)
+		assert.NoError(s.T(), err)
+		assert.NotNil(s.T(), order)
+		assert.Equal(s.T(), req.Qty, order.Qty)
+		assert.Zero(s.T(), decimal.Decimal{}, "", req.Notional, order.Notional)
```
gnvk, line 396: This assert doesn't check anything. It checks that `decimal.Decimal{}` is zero (always true). The rest of the arguments are just for printing. 
```suggestion
		assert.True(s.T(), req.Notional.IsZero())
		assert.True(s.T(), order.Notional.IsZero())
```

alpaca/alpaca_test.go
```diff
@@ -372,24 +372,45 @@ func (s *AlpacaTestSuite) TestAlpaca() {
...
+			Side:        Buy,
+			TimeInForce: GTC,
+			Type:        Limit,
+		}
+
+		order, err = PlaceOrder(req)
+		assert.NoError(s.T(), err)
 		assert.NotNil(s.T(), order)
+		assert.Equal(s.T(), req.Notional, order.Notional)
+		assert.Zero(s.T(), decimal.Decimal{}, "", req.Qty, order.Qty)
```
gnvk, line 413: Same comment

examples/mean-reversion/mean-reversion.go
```diff
@@ -199,12 +199,13 @@ func (alp alpacaClientContainer) rebalance() {
 func (alp alpacaClientContainer) submitLimitOrder(qty int, symbol string, price float64, side string) error {
 	account, _ := alp.client.GetAccount()
 	if qty > 0 {
+		qty := decimal.NewFromFloat(float64(qty))
```
gnvk, line 202: [minor] no need for this change

alpaca/alpaca_test.go
```diff
@@ -372,24 +372,45 @@ func (s *AlpacaTestSuite) TestAlpaca() {
...
 					Qty:         por.Qty,
+					Notional:    por.Notional,
 					Side:        por.Side,
 					TimeInForce: por.TimeInForce,
 					Type:        por.Type,
 				}),
 			}, nil
 		}
 
+		qty := decimal.New(1, 0)
```
gnvk, line 383: [minor] no need for this change

alpaca/rest.go
```diff
@@ -1177,3 +1178,29 @@ func unmarshal(resp *http.Response, data interface{}) error {
 
 	return json.Unmarshal(body, data)
 }
+
+type localPlaceOrderRequest PlaceOrderRequest
```
gnvk, line 1182: Please add a comment for this. It took me a while to figure out why it was needed.

alpaca/rest.go
```diff
@@ -1177,3 +1178,29 @@ func unmarshal(resp *http.Response, data interface{}) error {
...
+	}
+
+	// unmarshal to allow updates
+	data := make(map[string]interface{})
+	if err = json.Unmarshal(buf, &data); err != nil {
+		return nil, err
+	}
+
+	// remove zero-value struct fields related to order size
+	if req.Notional == reflect.Zero(reflect.TypeOf(req.Notional)).Interface() {
```
gnvk, line 1198: Is it okey to use `req.Notional.IsZero()` instead of reflection? Or am I missing something here?

alpaca/rest.go
```diff
@@ -1177,3 +1178,29 @@ func unmarshal(resp *http.Response, data interface{}) error {
 
 	return json.Unmarshal(body, data)
 }
+
+type localPlaceOrderRequest PlaceOrderRequest
+
+func (req PlaceOrderRequest) MarshalJSON() ([]byte, error) {
```
gnvk, line 1184: This is very hacky, but I couldn't find a better solution either with the standard json lib. Let's go with it :D

gnvk (commented):
Added some minor comments but looks good to me in general. Thank you for fixing the compatibility issue!

alpaca/rest.go
Neal: Does the SDK do any field validation check? If not, then just pass the request as is to the API and let it return?

But if you want some validations on client side, then just check if only one of qty or notional is non-zero? Ideally we could make them both pointers so we could do a more explicit nil check but that would be a breaking change.

alpaca/rest.go
Neal: Also note that you can add the `omitempty` json tag to qty and notional to not include the zero value in the json, which I believe should make this custom marshaller not necessary? Or am I missing something?

examples/mean-reversion/mean-reversion.go
Neal: I wonder why we cast the int to float64. I believe it can safely be updated to:

```go
decimal.NewFromInt(qty)
```

alpaca/rest.go
gnvk: Unfortunately it doesn't work: the zero value of `decimal.Decimal` is marshalled to 0.

alpaca/rest.go
gnvk: The only reason for this hack is to maintain backward compatibility in the exported structs.

alpaca/rest.go
Neal: oh right, that makes sense why this is required. in that case, I also agree `req.Notional.IsZero()` should work here instead of reflection I believe.

alpaca/alpaca_test.go
ducille: yup, forgot to clean this up after reverting the changes that made `Qty` a pointer

examples/mean-reversion/mean-reversion.go
ducille: Ironically, `decimal.NewFromInt()` doesn't actually accept `int` arguments: it expects `int64`. Since `rebalance()` declares `qty` to be an `int`, we actually need to cast it to something `decimal` can handle

alpaca/rest.go
ducille: added!

examples/mean-reversion/mean-reversion.go
Neal: aah fair enough, but wouldn't it still make sense to cast to int64 instead then and not deal with floating points?

ducille: yeah, you're right. it looks like this implements a discrete mean reversion strategy anyway, so no need to introduce floats here

gnvk (approved):
LGTM. Thanks @ducille!

PR #123, opened 2021-04-12
Add reworked data v2 stream client

kalmant:
Things left to do:
- Add an example after the dust has settled on the API

Update: the example has been added


v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+			go c.connWriter(ctx, &wg, closeCh)
+			wg.Wait()
+			c.logger.Warnf("datav2stream: connection lost")
+		}
+	}
+}
+
+var newPingTicker = func() ticker {
+	return &timeTicker{ticker: time.NewTicker(pingPeriod)}
+}
```
leki75, line 220: As this is only used in connPinger, why do not we put it there?

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+				c.connErrChan <- err
+			}
+			return
+		default:
+			if c.reconnectLimit != 0 && failedAttemptsInARow >= c.reconnectLimit {
+				c.logger.Errorf("datav2stream: max reconnect limit has been reached, last error: %v", connError)
+				c.connErrChan <- fmt.Errorf("max reconnect limit has been reached, last error: %w", connError)
+				return
+			}
+			time.Sleep(time.Duration(failedAttemptsInARow) * c.reconnectDelay)
```
leki75, line 176: minor: I know that this works but a little bit strange to me. `time.Duration(4)` means `4ns` and we multiply it with `c.reconnectDelay`.

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+			c.logger.Infof("datav2stream: connecting to %s, attempt %d/%d ...", u, failedAttemptsInARow, c.reconnectLimit)
+			conn, err := connCreator(ctx, u)
+			if err != nil {
+				connError = err
+				c.logger.Warnf("datav2stream: failed to connect, error: %v", err)
+				continue
+			}
+			c.conn = conn
+
+			c.logger.Infof("datav2stream: established connection")
```
leki75, line 187: minor: `c.logger.Info`?

v2/stream/new/flow.go
```diff
@@ -0,0 +1,193 @@
...
+	var retryErr error
+	// If we run into a retriable error during auth we should try
+	// again after a delay
+	for i := 0; i < authRetryCount; i++ {
+		if i > 0 {
+			n := i
+			if n > 10 {
+				n = 10
+			}
+			sleepDuration := authRetryDelayMultiplier * time.Duration(n*n)
```
leki75, line 39: The `authRetryDelayMultiplier` name or type is misleading.

v2/stream/new/client.go
leki75: Now, I see that we have a separate `Logger` interface only with `Infof`, `Warnf`, and `Errorf` methods.

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+// and handles communication both ways.
+//
+// After constructing, Connect() must be called before any subscription changes
+// are called. Connect keeps the connection alive and reestablishes it until
+// a configured number of retries has not been exceeded.
+//
+// Error() returns a channel that the client sends an error to when it has terminated.
+// A client can not be reused once it has terminated!
+//
+// SubscribeTo... and UbsubscribeFrom... can be used to modify subscriptions and
```
gnvk, line 22: typo: UnsubscribeFrom

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+// the handler used to process incoming trades/quotes/bars. These block until an
+// irrecoverable error occurs or if they succeed.
+//
+// Note that subscription changes can not be called concurrently.
+type StreamV2Client interface {
+	Connect(ctx context.Context) error
+	Error() <-chan error
```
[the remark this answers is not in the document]
gnvk: If this channel can only ever contain a single error, I would prefer this signature:
```suggestion
	Error() error
```

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+// Error() returns a channel that the client sends an error to when it has terminated.
+// A client can not be reused once it has terminated!
+//
+// SubscribeTo... and UbsubscribeFrom... can be used to modify subscriptions and
+// the handler used to process incoming trades/quotes/bars. These block until an
+// irrecoverable error occurs or if they succeed.
+//
+// Note that subscription changes can not be called concurrently.
+type StreamV2Client interface {
+	Connect(ctx context.Context) error
```
gnvk, line 28: I would copy the method docstrings to the interface. Or only have them here.

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+	quoteHandler          func(quote Quote)
+	bars                  []string
+	barHandler            func(bar Bar)
+	pendingSubChangeMutex sync.Mutex
+	pendingSubChange      *subChangeRequest
+}
+
+var _ StreamV2Client = (*client)(nil)
+
+func NewClient(feed string, opts ...Option) StreamV2Client {
```
gnvk, line 70: [minor] should be commented

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	switch ub.Scheme {
+	case "http", "ws":
+		scheme = "ws"
+	}
+
+	return url.URL{Scheme: scheme, Host: ub.Host, Path: "/v2/" + feed}, nil
+}
+
+var ErrConnectCalledMultipleTimes = errors.New("Connect called multiple times")
```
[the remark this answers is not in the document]
gnvk: 👍 for lowercase, and should be docstringed

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+
+// Connect establishes a connection for c using its configuration and reestablishes it when errors occur.
```
gnvk, line 102: Somehow I would try to highlight that it also keeps the connection open in the background forever.

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+func (c *client) Connect(ctx context.Context) error {
+	err := ErrConnectCalledMultipleTimes
+	c.connectOnce.Do(func() {
+		err = c.connect(ctx)
+		if err != nil {
+			c.connErrChan <- err
+		}
+		c.connectCalled = true
+	})
+	return err
```
gnvk, line 117: Probably I'm missing something. Isn't this return `ErrConnectCalledMultipleTimes` even on the first call?

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+
+	select {
+	case <-successCh:
+		return nil
+	case err := <-c.connErrChan:
+		return err
+	}
+}
+
+func (c *client) Error() <-chan error {
```
gnvk, line 137: See my comment on the interface

v2/stream/new/conn.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"context"
+	"time"
+)
+
+// conn represents a websocket connection between the server and the client
+type conn interface {
+	// close closes the websocket connection
+	close() error
```
gnvk, line 11: [minor] I would have the methods "exported". 

Reasoning: I agree that the interface itself should not be exported. However, if you wanted to use an external package that implements this interface, you couldn't. I actually can't think of a scenario when an interface method should be private.

v2/stream/new/flow.go
```diff
@@ -0,0 +1,193 @@
...
+		return err
+	}
+
+	return nil
+}
+
+var ErrNoConnected = errors.New("did not receive connected message")
+
+func (c *client) readConnected(ctx context.Context) error {
+	b, err := c.conn.readMessage(ctx)
```
gnvk, line 83: Is this not called concurrently with another read in `client.connReader`?

https://pkg.go.dev/nhooyr.io/websocket#Conn
> All methods may be called concurrently except for Reader and Read.

v2/stream/new/flow.go
```diff
@@ -0,0 +1,193 @@
...
+		return err
+	}
+	if len(resps) != 1 {
+		return ErrSubResponse
+	}
+
+	if resps[0].T == "error" {
+		return fmt.Errorf("sub: error from server: %s", resps[0].Msg)
+	}
+	if resps[0].T != "subscription" {
```
gnvk, line 185: What happens if we get data (for example a trade from a previous subscription) instead of the subscription confirmation here?

v2/stream/new/flow_test.go
```diff
@@ -0,0 +1,710 @@
+package new
```
gnvk, line 1: These tests are really good! 🥇

v2/stream/new/client.go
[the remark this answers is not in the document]
kalmant: My reasoning behind making this exported was that this way, users could actually expect/check for certain errors. If you think that's a bad idea I can make it unexported.
I'll add the docstrings either way.

v2/stream/new/log.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+type stdLog struct {
+	logger *log.Logger
+}
+
+var _ Logger = (*stdLog)(nil)
+
+func (s *stdLog) Infof(format string, v ...interface{}) {
+	// NOTE: there is no concept of levels in log
```
gnvk, line 21: So basically using the standard logger we only print errors, right?

v2/stream/new/options.go
```diff
@@ -0,0 +1,144 @@
...
+	barHandler     func(Bar)
+}
+
+func (o *options) apply(opts ...Option) {
+	for _, opt := range opts {
+		opt.apply(o)
+	}
+}
+
+type Option interface {
```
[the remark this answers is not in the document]
gnvk: Disagreed, `Option` is better in my opinion.

v2/stream/new/options.go
```diff
@@ -0,0 +1,144 @@
...
+		trades:         []string{},
+		tradeHandler:   func(t Trade) {},
+		quotes:         []string{},
+		quoteHandler:   func(q Quote) {},
+		bars:           []string{},
+		barHandler:     func(b Bar) {},
+	}
+}
+
+func (c *client) configure(o options) {
```
gnvk, line 75: [minor] this should go to client.go

v2/stream/new/options.go
```diff
@@ -0,0 +1,144 @@
...
+	c.barHandler = o.barHandler
+}
+
+func WithLogger(logger Logger) Option {
+	return newFuncOption(func(o *options) {
+		o.logger = logger
+	})
+}
+
+func WithHost(host string) Option {
```
gnvk, line 99: [minor] It would be nice to document these, although most of them are quite self-explanatory.

v2/stream/new/subscription.go
```diff
@@ -0,0 +1,142 @@
...
+import (
+	"errors"
+
+	"github.com/vmihailenco/msgpack/v5"
+)
+
+var ErrSubChangeBeforeConnect = errors.New("subscription change attempted before connecting")
+var ErrSubChangeAfterTerminated = errors.New("subscription change after client termination")
+var ErrSubChangeAlreadyInProgress = errors.New("subscription change already in progress")
+var ErrSubChangeInterrupted = errors.New("sub change interrupted by client termination")
```
gnvk, line 12: [minor] all above errors spell out "subscription". Either do the same or change them too. I prefer the former.

gnvk (commented):
🥇🥇🥇 Outstanding work! 🥇🥇🥇

v2/stream/new/client.go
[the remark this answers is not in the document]
kalmant: 1. The channel is buffered so "nothing bad" happens if the user does not read from it
2. If it were not a channel then users could not "listen" to it. They would have to poll it

v2/stream/new/client.go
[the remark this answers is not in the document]
gnvk: You misunderstood: it should remain exported, the error string should be lowercase!

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+	ub, err := url.Parse(host)
+	if err != nil {
+		return url.URL{}, err
+	}
+	switch ub.Scheme {
+	case "http", "ws":
+		scheme = "ws"
+	}
+
+	return url.URL{Scheme: scheme, Host: ub.Host, Path: "/v2/" + feed}, nil
```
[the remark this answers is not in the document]
kalmant: I was following what the old client used to do. I can make this change if you think this would be beneficial.

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+
+	successCh := make(chan struct{})
+	go c.maintainConnection(ctx, u, successCh)
+
+	select {
+	case <-successCh:
+		return nil
+	case err := <-c.connErrChan:
+		return err
+	}
```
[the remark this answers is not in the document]
kalmant: You are right! This was not the case in a previous version but as things got changed this mistake did slip in. I'm fixing this ASAP!

v2/stream/new/client.go
kalmant: I'm not aware of any other way due to how types work in the language.

v2/stream/new/client.go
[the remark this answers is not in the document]
gnvk: We desperately need that example :P

v2/stream/new/client.go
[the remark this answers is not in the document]
kalmant: Fixed.

v2/stream/new/client.go
kalmant: Yes, I wanted to keep the interface as small as possible.

v2/stream/new/client.go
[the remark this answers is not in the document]
kalmant: 🤦🏼 
I think I've got it now

v2/stream/new/client.go
```diff
@@ -0,0 +1,336 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case msg, ok := <-c.in:
+			if !ok {
+				return
+			}
+			c.logger.Errorf("processing msg: %s", msg)
```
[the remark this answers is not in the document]
kalmant: well...

v2/stream/new/flow.go
kalmant: I am open to suggestions

v2/stream/new/client.go
kalmant: I'll give it a shot.

v2/stream/new/client.go
kalmant: No, because `err = c.connect(ctx)` overwrites that error.

v2/stream/new/conn.go
kalmant: Who is the 'you' in this case? Who tries to use an external package that implements the interface?

v2/stream/new/flow.go
kalmant: `flow.go`'s methods are only called by `initialize`.
`initialize` in turn is only called when `connReader` pinger is not running.

v2/stream/new/flow.go
kalmant: Although the naming may be a bit confusing, functions in `flow.go` are only called by `initialize`: this is when we connect to the server, authenticate and set up initial subscriptions.
In this flow no other messages are acceptable.

v2/stream/new/log.go
kalmant: Yes.
This can be changed if you think that's the way to go but I wasn't sure about printing infos/warns by default or using multiple loggers.

v2/stream/new/client.go
gnvk: As I said, I was **_probably_** missing something. No I can 🔫  myself.

v2/stream/new/options.go
kalmant: Done

v2/stream/new/flow.go
gnvk: You are right. My bad again. 🔫 🔫

v2/stream/new/flow.go
gnvk: It's clear now. Sorry. 🔫🔫🔫

v2/stream/new/client.go
kalmant: Done

v2/stream/new/options.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (fo *funcOption) apply(o *options) {
+	fo.f(o)
+}
+
+func newFuncOption(f func(*options)) *funcOption {
+	return &funcOption{
+		f: f,
+	}
+}
```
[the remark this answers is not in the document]
kalmant: Thank you :) I did not come up with this, gnvk helped and I think we both saw some great examples out there in the wild.

v2/stream/new/options.go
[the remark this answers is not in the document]
kalmant: Since this is a tie I'm not going to change it.

v2/stream/new/conn_nhoory.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// isCloseError returns true if the given error represents a standard websocket closure
+func (c *nhooyrWebsocketConn) isCloseError(err error) bool {
+	switch websocket.CloseStatus(err) {
+	case
+		websocket.StatusNormalClosure,
+		websocket.StatusGoingAway,
+		websocket.StatusNoStatusRcvd,
+		websocket.StatusAbnormalClosure:
```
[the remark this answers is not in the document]
kalmant: From https://tools.ietf.org/html/rfc6455:
   1005

      1005 is a reserved value and MUST NOT be set as a status code in a
      Close control frame by an endpoint.  It is designated for use in
      applications expecting a status code to indicate that no status
      code was actually present.

   1006

      1006 is a reserved value and MUST NOT be set as a status code in a
      Close control frame by an endpoint.  It is designated for use in
      applications expecting a status code to indicate that the
      connection was closed abnormally, e.g., without sending or
      receiving a Close control frame.
     
But (!) I think it makes sense to question whether these are all cases where we wouldn't log an error. I think they are not. What do you think?

v2/stream/new/client.go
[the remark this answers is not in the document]
kalmant: This will stay like this for now.

v2/stream/new/options.go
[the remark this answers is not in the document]
leki75: This is against Effective Go: https://golang.org/doc/effective_go#interface-names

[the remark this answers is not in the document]
kalmant: The reason this name/solution was picked because using the Go gRPC SDK is so "natural". As you can see [they are using](https://github.com/grpc/grpc-go/blob/master/dialoptions.go#L76) the exact same naming convention and I think it works very well. Especially because the only method is unexported.

v2/stream/new/options.go
```diff
@@ -0,0 +1,147 @@
...
+// WithCredentials configures the key and secret to use
+func WithCredentials(key, secret string) Option {
+	return newFuncOption(func(o *options) {
+		o.key = key
+		o.secret = secret
+	})
+}
+
+// WithReconnectSettings configures how many consecutive connection
+// errors should be accepted and the delay between retries
```
gnvk, line 100: Please add that the delay is actually multiplied by the consecutive error count!

gnvk (approved):
Outstanding work.

PR #125, opened 2021-04-20
Add latest trades and quotes

kalmant:


alpaca/alpaca_test.go
```diff
@@ -265,6 +266,106 @@ func (s *AlpacaTestSuite) TestAlpaca() {
...
+		}
+		do = func(c *Client, req *http.Request) (*http.Response, error) {
+			return &http.Response{
+				Body: ioutil.NopCloser(strings.NewReader(latestTradeJSON)),
+			}, nil
+		}
+
+		actualLatestTrade, err := GetLatestTrade("AAPL")
+		assert.NotNil(s.T(), actualLatestTrade)
+		assert.Nil(s.T(), err)
```
gnvk, line 305: [minor] `NoError` is preferable

gnvk: same for a few more places in this file

alpaca/alpaca_test.go
```diff
@@ -265,6 +266,106 @@ func (s *AlpacaTestSuite) TestAlpaca() {
 		assert.Nil(s.T(), actualLastTrade)
 	}
 
+	// get latest trade
```
gnvk, line 269: I'd prefer to move this to a separate test case but agree with you to stay consistent with the rest of the file

gnvk (approved):
LGTM

alpaca/alpaca_test.go
kalmant: You are right, I fixed it throughout the file.

alpaca/alpaca_test.go
kalmant: Yeah, I did not want to change things in this PR

PR #127, opened 2021-04-28
Remove polygon and old stream from v2

gnvk:


leki75 (approved):
I find the original example really complicated. This is a good step to make it more readable. Thank you so much!

PR #133, opened 2021-06-03
Add crypto stream

gnvk:


examples/crypto-stream/crypto_stream.go
```diff
@@ -0,0 +1,48 @@
...
+		stream.WithLogger(&logger{}),
+		// TODO: REPLACE!
+		stream.WithBaseURL("http://localhost:8080/v1alpha1/crypto"),
+		// configuring initial subscriptions and handlers
+		stream.WithCryptoTrades(func(ct stream.CryptoTrade) {
+			fmt.Printf("TRADE %s %g of $%g at %s\n", ct.Symbol, ct.Size, ct.Price, ct.Timestamp)
+		}, "*"),
+		stream.WithCryptoQuotes(func(cq stream.CryptoQuote) {
+			fmt.Printf("QUOTE %s bid: $%g ask: $%g at %s\n", cq.Symbol, cq.BidPrice, cq.AskPrice, cq.Timestamp)
+		}, "AAPL", "BTCUSD"),
```
kalmant, line 26: Please remove `AAPL`, it's a bit confusing to see it in in a crypto client.

marketdata/stream/message.go
```diff
@@ -67,7 +75,15 @@ func (c *client) handleMessage(b []byte) error {
 	return nil
 }
 
-func (c *client) handleTrade(d *msgpack.Decoder, n int) error {
+type stocksMsgHandler struct {
```
kalmant, line 78: An interface compliance check would be nice for both `stockMsgHandler` and `cryptoMsgHandler`

marketdata/stream/client.go
```diff
@@ -65,63 +95,85 @@ type client struct {
...
 	c.dailyBars = o.dailyBars
-	c.dailyBarHandler = o.dailyBarHandler
+	c.connCreator = o.connCreator
+}
+
+type stocksClient struct {
+	*client
+
+	feed    string
+	handler *stocksMsgHandler
```
kalmant, line 138: I wish `handler` weren't handler because of the clash it causes with the embedded struct. Same for `cryptoClient`. I can't recommend a better name right now but perhaps you have a good idea.

marketdata/stream/message_test.go
```diff
@@ -126,6 +162,33 @@ var testBar = barWithT{
 	Timestamp: time.Date(2021, 03, 05, 16, 0, 0, 0, time.UTC),
 }
 
+var testCryptoTrade = cryptoTradeWithT{
+	Type:      "t",
+	Symbol:    "A",
+	Price:     100,
+	Size:      10,
```
kalmant, line 169: Please use floats here (same for `testCryptoBar`)

marketdata/stream/client_test.go
```diff
@@ -203,15 +284,15 @@ func TestSubscribeCalledButClientTerminates(t *testing.T) {
 	checkInitialMessagesSentByClient(t, connection, "", "", []string{}, []string{}, []string{}, []string{})
 	subErrCh := make(chan error, 1)
 	subFunc := func() {
-		subErrCh <- c.SubscribeToTrades(func(trade Trade) {}, "ALPACA")
+		subErrCh <- c.SubscribeToTrades(func(trade CryptoTrade) {}, "PACOIN")
```
kalmant, line 287: 🔮

kalmant: FYI There are a few places where `PACOIN` is not used for crypto trades/quotes.

marketdata/stream/message.go
gnvk: It's in `NewStocksClient()`.

marketdata/stream/client.go
gnvk: This was actually intentional. `sc.client.handler` is an interface pointing to `sc.handler`. `sc.client.handler` should never be used from a caller's perspective.

marketdata/stream/message.go
kalmant: Then one could argue that `var _ StocksClient = (*stocksClient)(nil)` and `var _ CryptoClient = (*cryptoClient)(nil)` is not "necessary" either because of the `New...Client()` functions.
I'm fine with either solution but let's stick to one.

kalmant (approved):
All of my outstanding comments are minor things so ✅ 
Great job!

marketdata/stream/message.go
gnvk: 👍  Added it for consistency.

PR #137, opened 2021-07-16
Logging improvements

gnvk:


marketdata/stream/flow.go
```diff
@@ -55,6 +55,7 @@ func (c *client) initialize(ctx context.Context) error {
 		if !isErrorRetriable(retryErr) {
 			return retryErr
 		}
+		c.logger.Infof("datav2stream: auth error: %s, will retry", retryErr)
```
kalmant, line 58: There is some logging on L40 that is similar to this one. Perhaps here you could omit the "will retry" part of the message and let the other one log that information (if it actually retries)
Oh and please fix the typo on L40 (retring)

gnvk: I fixed the typo. I find it useful to log it on both places.

kalmant: but here you might log `... will retry` even though there won't be any more retries

marketdata/stream/log.go
```diff
@@ -2,40 +2,44 @@ package stream
...
-var _ Logger = (*stdLog)(nil)
+type defaultLogger struct{}
 
-func (s *stdLog) Infof(format string, v ...interface{}) {
-	// NOTE: there is no concept of levels in the stdlib log package
-	// For less noise, this default implementation will not print any non-error messages
-	// To see these write your own implementation or use a proper logger,
-	//  e.g. https://github.com/uber-go/zap
+func (*defaultLogger) Infof(format string, v ...interface{}) {
+	log.Printf("INFO "+format, v...)
```
kalmant, line 17: Would adding a colon after the level be a bad idea?

gnvk: I don't like it, it looks better without it in my opinion.

marketdata/stream/flow.go
gnvk: You convinced me, I remove it.

kalmant (approved):
The two remaining comments are minor so feel free to merge if you don't think those comments are important.

marketdata/stream/flow.go
```diff
@@ -149,21 +154,22 @@ func (c *client) readAuthResponse(ctx context.Context) error {
...
 			return ErrInvalidCredentials
+		case 406:
+			// A previous connection may be "stuck" on the server so we may run into
+			// `[{"T":"error","code":406,"msg":"connection limit exceeded"}]`
+			return ErrConnectionLimitExceeded
+		case 409:
+			return ErrInsufficientSubscription
 		}
-		return err
+		return errors.New(resp.Msg)
```
leki75, line 170: This is switch default, right?

marketdata/stream/flow.go
```diff
@@ -127,6 +128,10 @@ var ErrConnectionLimitExceeded = errors.New("connection limit exceeded")
 // ErrInvalidCredentials is returned when invalid credentials have been sent by the user
 var ErrInvalidCredentials = errors.New("invalid credentials")
 
+// ErrInsufficientSubscription is returned when the user does not have proper
+// subscription for the requested feed (e.g. SIP)
+var ErrInsufficientSubscription = errors.New("insufficient subscription")
+
```
leki75, line 134: It is a little bit weird that we define `isErrorIrrecoverable()` in `client.go` while `isErrorRetriable()` here.

leki75 (commented):
LGTM

marketdata/stream/flow.go
gnvk: It's exactly the same, yes.

marketdata/stream/flow.go
gnvk: My guess is that it was originally put there because of the usage.

PR #141, opened 2021-07-30
Add ADTV and SMA

kalmant:
I'm not sure whether it makes sense to include the number of days in `marketdata.ADTV`

### DIFF
--- a/alpaca/rest.go
+++ b/alpaca/rest.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


alpaca/rest.go
```diff
@@ -703,6 +704,72 @@ func (c *Client) GetSnapshots(symbols []string) (map[string]*marketdata.Snapshot
...
+		Volume: float64(totalVolume) / float64(count),
+		Days:   count,
+	}, nil
+}
+
+// GetSMA returns the Simple Moving Average (SMA) starting at start, ending at end
+// and using window as the averaging period.
+//
+// Deprecated: will be moved to the marketdata package!
+func (c *Client) GetSMA(symbol string, start, end time.Time, window int) ([]marketdata.ValueAtTime, error) {
```
leki75, line 736: This function has limited usage: eg. I could have SMA for volumes, opening/closing prices, or any other gauge data that I like. I could count all of them just to querying the underlying endpoint just once. If we implement everything passing all the information as parameters then the function signature will be too complex.

kalmant:
We might want to reopen this so don't delete the branch!

PR #154, opened 2021-09-30
Fix marketdata stream subscribe errors

gnvk:


marketdata/stream/subscription.go
```diff
@@ -177,7 +187,13 @@ func (c *client) handleSubChange(
...
 	}
 
-	return <-request.result
+	select {
+	case err := <-request.result:
+		return err
+	case <-timeAfter(3 * time.Second):
+	}
+
+	return ErrSubscriptionChangeTimeout
```
kalmant, line 196: How would a user deal with this error? Would they retry the call? Wouldn't that result in a `ErrSubscriptionChangeAlreadyInProgress` error?

marketdata/stream/client_test.go
```diff
@@ -327,6 +327,72 @@ func TestSubscribeCalledButClientTerminatesCrypto(t *testing.T) {
...
+	}
+
+	go subFunc()
+	mockTimeAfterCh <- time.Now()
+	err = <-subErrCh
+	assert.Error(t, err)
+	assert.True(t, errors.Is(err, ErrSubscriptionChangeTimeout), "actual: %s", err)
+}
+
+func TestSubscriptionChangeInvalid(t *testing.T) {
```
kalmant, line 364: Great test :)

marketdata/stream/subscription.go
gnvk: Excellent note, thanks.

PR #157, opened 2021-10-06
Consider all HTTP 4xx irrecoverable errors

gnvk:
Currently if on the marketdata stream we receive an HTTP 422 from the server, the client will try to reconnect. This makes no sense, because the response will always be the same.

### DIFF
--- a/marketdata/stream/client.go
+++ b/marketdata/stream/client.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


marketdata/stream/client.go
```diff
@@ -379,6 +383,16 @@ func isErrorIrrecoverable(err error) bool {
 	return errors.Is(err, ErrInvalidCredentials) || errors.Is(err, ErrInsufficientSubscription)
 }
 
+func isHttp422(err error) bool {
+	// Unfortunately the nhoory error is a simple formatted string, created by fmt.Errorf,
+	// so the only check we can do is string matching
+	return strings.Contains(err.Error(), "expected handshake response status code 101 but got 422")
```
kalmant, line 389: Yes, this is pretty unfortunate

PR #168, opened 2021-10-25
Add ADTV to the marketdata examples

gnvk:


examples/marketdata/marketdata.go
```diff
@@ -50,4 +50,35 @@ func main() {
...
+func getADTV(symbol string, start, end time.Time) (av float64, n int, err error) {
+	var (
+		totalVolume uint64
+	)
+	for item := range marketdata.GetBarsAsync(symbol, marketdata.GetBarsParams{
+		Start: start,
+		End:   end,
+	}) {
+		if err = item.Error; err != nil {
+			return
```
kalmant, line 74: I think a panic or printing an error or something would be prudent here

gnvk: We return the error and the caller will panic.

kalmant: You are right, my bad

kalmant (approved):
Thank you for adding this example.

PR #175, opened 2021-11-23
Add internal received at

gnvk:


marketdata/stream/entities.go
```diff
@@ -12,6 +12,17 @@ type Trade struct {
...
 	Tape       string
+
+	internal tradeInternal
+}
+
+type tradeInternal struct {
+	ReceivedAt time.Time
+}
+
+// Internal contains internal fields.
```
kalmant, line 23: Please expand on this comment: "Do not expect any of the fields to be set and no future compatibility is promised for this struct. You should not use this at all" or something similar. The current comment is not scary enough.

kalmant (approved):
Perfect

PR #177, opened 2021-12-06
Handle trade error cancels and corrections

ccnlui:
## Context
Handle trade cancel errors and correction messages, which are subscribed together with trades automatically.

## TODO
Test tomorrow.


marketdata/stream/client.go
```diff
@@ -53,6 +53,10 @@ type StocksClient interface {
 	UnsubscribeFromStatuses(symbols ...string) error
 	SubscribeToLULDs(handler func(luld LULD), symbols ...string) error
 	UnsubscribeFromLULDs(symbols ...string) error
+	RegisterHandlerCancelErrors(handler func(tce TradeCancelError))
```
gnvk, line 56: [minor] Maybe I would skip the "Handler" from the method name. Not completely sure though.

marketdata/stream/options.go
```diff
@@ -233,6 +237,20 @@ func WithLULDs(handler func(LULD), symbols ...string) StockOption {
 	})
 }
 
+// WithCancelErrors configures inital trade cancel errors handler
```
gnvk, line 240: Please extend the docstring with the fact that the symbols are missing from the option, because ...

gnvk (commented):
Please also update the unit test suite with the new message types! Otherwise the PR looks solid.

marketdata/stream/client.go
ccnlui: :thinking: Removed to make it cleaner

ccnlui:
Added test cases to check error cancels and corrections, rebase to master as well.

gnvk (approved):
Great work @ccnlui !

Issue #176, opened 2021-11-30
Not possible to list non-trade Account activities


JasonTashtego:
This struct in entities.go defines the Date field as time.Time,  but the value (from Alpaca) is of the form 'YYYY-MM-dd'  for non-trade activities so it always fails to unmarshall the data from Alpaca.   For example I have a 'SPIN' activity from a spin off.

```
type AccountActivity struct {
	ID              string          `json:"id"`
	ActivityType    string          `json:"activity_type"`
	TransactionTime time.Time       `json:"transaction_time"`
	Type            string          `json:"type"`
	Price           decimal.Decimal `json:"price"`
	Qty             decimal.Decimal `json:"qty"`
	Side            string          `json:"side"`
	Symbol          string          `json:"symbol"`
	LeavesQty       decimal.Decimal `json:"leaves_qty"`
	CumQty          decimal.Decimal `json:"cum_qty"`
	Date            time.Time       `json:"date"`
	NetAmount       decimal.Decimal `json:"net_amount"`
	Description     string          `json:"description"`
	PerShareAmount  decimal.Decimal `json:"per_share_amount"`
}
```

Bexanderthebex:
@JasonTashtego attempted to fix this on this PR. Sorry for the late action.

PR #178, opened 2021-12-07
ENG-8813: Fix wrong `Date` data type for NTAs

Bexanderthebex:
Changes:

- Fixed the issue raised over here: https://github.com/alpacahq/alpaca-trade-api-go/issues/176
- Added cloud.google.com/go package
- Added unit test


alpaca/entities.go
```diff
@@ -159,7 +160,7 @@ type AccountActivity struct {
 	Symbol          string          `json:"symbol"`
 	LeavesQty       decimal.Decimal `json:"leaves_qty"`
 	CumQty          decimal.Decimal `json:"cum_qty"`
-	Date            time.Time       `json:"date"`
+	Date            civil.Date      `json:"date"`
```
gnvk, line 163: This is technically a backward-incompatible change. However, if the original code never worked (??? is this really the case??) and the new one works with all possible account activities that the server may send, then it could be fine.

alpaca/entities_test.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"encoding/json"
+	"testing"
+
+	"cloud.google.com/go/civil"
+	"github.com/shopspring/decimal"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNonTradeActivities(t *testing.T) {
```
gnvk, line 12: First of all: thank you for adding a unit test! But, could you please create a test similar to the other one in rest_test.go? Other than keeping the consistency in the repo it would also be useful to test the whole `GetAccountActivities` method, not only the unmarshaling.

alpaca/entities.go
Bexanderthebex: The original code in the sdk never worked. Server-side code never changed.

gnvk (approved):
LGTM. Thank you for fixing this!

PR #184, opened 2022-01-19
Mark v1 marketdata endpoints deprecated

gnvk:


leki75 (commented):
Shouldn't we deprecate polygon.io endpoints, too?
Examples of long-short and mean-revision are using the deprecated v1 API.

gnvk:
> Shouldn't we deprecate polygon.io endpoints, too? 

Good point, I did.

> Examples of long-short and mean-revision are using the deprecated v1 API.

You are right, but I think it isn't worth the effort to change them in the v1 branch. In v2, they're using the v2 methods.

leki75 (approved):
Thank you!

PR #190, opened 2022-02-15
Move updated bars to their own subscription

gnvk:


marketdata/stream/options.go
```diff
@@ -295,19 +307,21 @@ func defaultCryptoOptions() *cryptoOptions {
...
 			},
 		},
-		tradeHandler:    func(t CryptoTrade) {},
-		quoteHandler:    func(q CryptoQuote) {},
-		barHandler:      func(b CryptoBar) {},
-		dailyBarHandler: func(b CryptoBar) {},
+		tradeHandler:      func(t CryptoTrade) {},
+		quoteHandler:      func(q CryptoQuote) {},
+		barHandler:        func(b CryptoBar) {},
+		updatedBarHandler: func(cb CryptoBar) {},
```
ccnlui, line 323: should this be `b` instead of `cb`?

ccnlui:
Similar to above, this should be updated to include `updatedBars` I think.
https://github.com/alpacahq/alpaca-trade-api-go/blob/782ba398e018e123652d6af7910bf0bb53cabf77/marketdata/stream/message.go#L773

ccnlui (approved):

PR #193, opened 2022-03-09
Reuse the HTTP client and the HTTP connections

gnvk:


alpaca/rest.go
```diff
@@ -81,8 +84,15 @@ func NewClient(opts ClientOpts) Client {
 	if opts.RetryDelay == 0 {
 		opts.RetryDelay = time.Second
 	}
+	httpClient := opts.HttpClient
+	if httpClient == nil {
+		httpClient = &http.Client{
+			Timeout: opts.Timeout,
```
leki75, line 90: What if the client provides both options: HttpClient and Timeout?

leki75 (approved):
LGTM

alpaca/rest.go
gnvk: Good question. The HttpClient has precedence. I'll add a comment.

Issue #108, opened 2021-01-08
Add missing params to ListOrders


coopernurse:
ListOrders currently supports a subset of the available params.  Missing params:

* after
* direction
* symbols

I'd suggest we add a new method that accepts a request struct that includes all supported params.  I'm open to naming suggestions.  We could deprecate the old method.  The result would be something like:

```
func (c *Client) ListOrders(status *string, until *time.Time, limit *int, nested *bool) ([]Order, error) {
	req := ListOrdersRequest{
		Status: status,
		Until:  until,
		Limit:  limit,
		Nested: nested,
	}
	return c.ListOrders2(req)
}

func (c *Client) ListOrders2(req ListOrdersRequest) ([]Order, error) {
// code here
}
```

gnvk:
Closed by https://github.com/alpacahq/alpaca-trade-api-go/commit/91b047240e0c1a4fd9726c699d0fdd405435decf

PR #196, opened 2022-03-23
Feature/list orders

repositoryFixer:
In this PR I added the ListOrdersWithRequest function because the ListOrders function that is already present does not include all possible parameters and fixing this directly in the function would be a breaking change.

As requested by this issue: https://github.com/alpacahq/alpaca-trade-api-go/issues/108

### DIFF
--- a/alpaca/rest.go
+++ b/alpaca/rest.go
@@ -29,7 +29,7 @@ type Client interface {
 	GetClock() (*Clock, error)
 	GetCalendar(start, end *string) ([]CalendarDay, error)
 	ListOrders(status *string, until *time.Time, limit *int, nested *bool) ([]Order, error)
-	ListOrdersV2(req ListOrderRequest) ([]Order, error)
+	ListOrdersV2(req ListOrdersRequest) ([]Order, error)
 	PlaceOrder(req PlaceOrderRequest) (*Order, error)
 	GetOrder(orderID string) (*Order, error)
 	GetOrderByClientOrderID(clientOrderID string) (*Order, error)
@@ -473,7 +473,7 @@ func (c *client) ListOrders(status *string, until *time.Time, limit *int, nested
 
 // ListOrdersV2 returns the list of orders for an account,
 // filtered by the input parameters.
-func (c *client) ListOrdersV2(req ListOrderRequest) ([]Order, error) {
+func (c *client) ListOrdersV2(req ListOrdersRequest) ([]Order, error) {
 	urlString := fmt.Sprintf("%s/%s/orders", c.opts.BaseURL, apiVersion)
 	if req.Nested != nil {
 		urlString += fmt.Sprintf("?nested=%v", *req.Nested)
--- a/alpaca/rest.go
+++ b/alpaca/rest.go
@@ -29,6 +29,7 @@ type Client interface {
 	GetClock() (*Clock, error)
 	GetCalendar(start, end *string) ([]CalendarDay, error)
 	ListOrders(status *string, until *time.Time, limit *int, nested *bool) ([]Order, error)
+	ListOrdersV2(req ListOrderRequest) ([]Order, error)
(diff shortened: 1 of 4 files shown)


alpaca/rest.go
```diff
@@ -29,6 +29,7 @@ type Client interface {
 	GetClock() (*Clock, error)
 	GetCalendar(start, end *string) ([]CalendarDay, error)
 	ListOrders(status *string, until *time.Time, limit *int, nested *bool) ([]Order, error)
+	ListOrdersV2(req ListOrdersRequest) ([]Order, error)
```
gnvk, line 32: I don't like the naming. What do you think about `ListOrdersWithRequest`?

alpaca/rest.go
```diff
@@ -429,6 +430,7 @@ func (c *client) GetCalendar(start, end *string) ([]CalendarDay, error) {
 
 // ListOrders returns the list of orders for an account,
 // filtered by the input parameters.
+// Deprecated: This function is deprecated in favor of ListOrdersV2 which contains all possible parameters.
```
gnvk, line 433: [minor] Please add a newline before the deprecation notice!

alpaca/rest.go
```diff
@@ -469,6 +471,60 @@ func (c *client) ListOrders(status *string, until *time.Time, limit *int, nested
 	return orders, nil
 }
 
+// ListOrdersV2 returns the list of orders for an account,
+// filtered by the input parameters.
+func (c *client) ListOrdersV2(req ListOrdersRequest) ([]Order, error) {
+	urlString := fmt.Sprintf("%s/%s/orders", c.opts.BaseURL, apiVersion)
+	if req.Nested != nil {
```
gnvk, line 478: [minor] Why is Nested handled differently than all the other parameters?

alpaca/rest_test.go
```diff
@@ -171,6 +171,36 @@ func TestListOrders(t *testing.T) {
 	assert.Nil(t, orders)
 }
 
+func TestListOrdersV2(t *testing.T) {
```
gnvk, line 174: [minor] This is a pretty weak test. I understand the same is true for pretty much all the other test cases in this file, but it would be nice to improve it. A good test would cover at least two cases: one with an "empty" request, and one with all, or almost all possible options, both with assertions on the request. Please see marketdata/rest_test.go for examples.

alpaca/rest.go
```diff
@@ -469,6 +471,60 @@ func (c *client) ListOrders(status *string, until *time.Time, limit *int, nested
...
+	if req.After != nil {
+		q.Set("after", req.After.Format(time.RFC3339))
+	}
+
+	if req.Until != nil {
+		q.Set("until", req.Until.Format(time.RFC3339))
+	}
+
+	if req.Limit != nil {
+		q.Set("limit", strconv.FormatInt(int64(*req.Limit), 10))
```
gnvk, line 501: [minor] 
```suggestion
		q.Set("limit", strconv.Itoa(*req.Limit))
```

alpaca/rest.go
repositoryFixer: That's fine with me.

alpaca/rest.go
repositoryFixer: Done

alpaca/rest.go
repositoryFixer: This is a relict from the other ListOrders function. I fixed it so nested is handled the same as all the other parameters.

alpaca/rest_test.go
repositoryFixer: I changed the test to also check if the query params are set correctly.

alpaca/rest.go
repositoryFixer: Done

gnvk (approved):
Thank you for fixing this.

PR #208, opened 2022-06-30
Add GetOrderWithRequest to allow getting nested orders

probably-not:
According to the Alpaca docs, the [Get an Order](https://alpaca.markets/docs/api-references/trading-api/orders/#get-an-order) endpoint allows requesting nested orders, so that we can get all the legs of an advanced order. The current SDK does not allow this. This PR fixes that by adding a `GetOrderRequest` struct and a `GetOrderWithRequest` function to the client that allows specifying values on the request to get an order (for now, only the `Nested *bool` value).

### DIFF
--- a/alpaca/entities.go
+++ b/alpaca/entities.go
@@ -244,6 +244,11 @@ type ListOrdersRequest struct {
 	Symbols   *string    `json:"symbols"`
 }
 
+type GetOrderRequest struct {
+	ID     string `json:"id"`
+	Nested *bool  `json:"nested"`
+}
+
 type Side string
 
 const (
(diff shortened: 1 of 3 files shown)


gnvk (commented):
The PR is great 👍  However, unfortunately the server doesn't seem to handle the `nested=false` case properly. Let's wait until that's fixed before merging this to avoid confusion.

probably-not:
Thanks @gnvk !

probably-not:
Hey @gnvk, any movement on the backend bug on your end? We'd love to get this merged in so we can get back onto the official client and not the fork

gnvk:
> Hey @gnvk, any movement on the backend bug on your end? We'd love to get this merged in so we can get back onto the official client and not the fork

I have good and bad news. The good news is that you can use the official client and not the fork. The bad news is that I'm going to need to close this beautiful PR. The thing is that this parameter was never actually in place. It was a documentation bug, that should be fixed now. This means that you can simply use the `GetOrder` function to get your order with its legs.

Sorry for the inconvenience and I really really hope you can contribute to the SDK in other meaningful ways in the future.

PR #210, opened 2022-07-25
Add (dis)connect callbacks to market data stream client

dylanmctiernan-alpaca:
Allows the market data streaming client to supply a callback function as an option to run on connect/disconnect from the stream. The primary use case is for supporting monitoring and alerting workflows (e.g. prometheus metrics).

### DIFF
--- a/marketdata/stream/client.go
+++ b/marketdata/stream/client.go
@@ -107,8 +107,8 @@ type client struct {
 
 	reconnectLimit     int
 	reconnectDelay     time.Duration
-	connectCallback    func() error
-	disconnectCallback func() error
+	connectCallback    func()
+	disconnectCallback func()
 	processorCount     int
 	bufferSize         int
 	connectOnce        sync.Once
@@ -368,7 +368,7 @@ func (c *client) maintainConnection(ctx context.Context, u url.URL, initialResul
 		// If a disconnect/connect callback is running then wait until it finishes or times out.
 		timeout := time.Second
 		if waitTimeout(&callbackWaitGroup, timeout) {
-			c.logger.Warnf("datav2stream: timed out after waiting one second for connect/disconnect callbacks to return")
+			c.logger.Warnf("datav2stream: timed out after waiting %s for connect/disconnect callbacks to return", timeout)
 		}
 	}()
 
@@ -432,9 +432,7 @@ func (c *client) maintainConnection(ctx context.Context, u url.URL, initialResul
 				callbackWaitGroup.Add(1)
 				go func() {
 					defer callbackWaitGroup.Done()
-					if err := c.connectCallback(); err != nil {
-						c.logger.Warnf("datav2stream: connect callback failed, error: %v", err)
-					}
+					c.connectCallback()
 				}()
 			}
 
@@ -466,9 +464,7 @@ func (c *client) maintainConnection(ctx context.Context, u url.URL, initialResul
 				callbackWaitGroup.Add(1)
 				go func() {
 					defer callbackWaitGroup.Done()
-					if err := c.disconnectCallback(); err != nil {
-						c.logger.Warnf("datav2stream: disconnect callback failed, error: %v", err)
-					}
+					c.disconnectCallback()
 				}()
 			}
 		}
--- a/marketdata/stream/client.go
+++ b/marketdata/stream/client.go
@@ -105,17 +105,19 @@ type client struct {
 	key     string
 	secret  string
 
-	reconnectLimit int
-	reconnectDelay time.Duration
-	processorCount int
-	bufferSize     int
-	connectOnce    sync.Once
-	connectCalled  bool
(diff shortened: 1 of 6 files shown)


marketdata/stream/client.go
```diff
@@ -360,6 +365,11 @@ func (c *client) maintainConnection(ctx context.Context, u url.URL, initialResul
 		if connectedAtLeastOnce {
 			close(c.terminatedChan)
 		}
+		// If a disconnect/connect callback is running then wait until it finishes or times out.
+		timeout := time.Second
+		if waitTimeout(&callbackWaitGroup, timeout) {
+			c.logger.Warnf("datav2stream: timed out after waiting one second for connect/disconnect callbacks to return")
```
gnvk, line 371: [minor] If the `timeout` is already in a variable, it should be used in the log as well.

marketdata/stream/client.go
```diff
@@ -417,6 +427,17 @@ func (c *client) maintainConnection(ctx context.Context, u url.URL, initialResul
 				continue
 			}
 			c.logger.Infof("datav2stream: finished connection setup")
+
+			if c.connectCallback != nil {
+				callbackWaitGroup.Add(1)
+				go func() {
+					defer callbackWaitGroup.Done()
+					if err := c.connectCallback(); err != nil {
+						c.logger.Warnf("datav2stream: connect callback failed, error: %v", err)
```
gnvk, line 436: If we all do with the returned error of the callback functions is logging then I think we shouldn't do that. They shouldn't return an error and any error handling should be done by the user inside the functions.

marketdata/stream/options.go
```diff
@@ -148,14 +170,16 @@ func defaultStockOptions() *stockOptions {
...
-			reconnectDelay: 150 * time.Millisecond,
-			processorCount: 1,
-			bufferSize:     100000,
+			logger:             DefaultLogger(),
+			baseURL:            baseURL,
+			key:                os.Getenv("APCA_API_KEY_ID"),
+			secret:             os.Getenv("APCA_API_SECRET_KEY"),
+			reconnectLimit:     20,
+			reconnectDelay:     150 * time.Millisecond,
+			connectCallback:    nil,
```
gnvk, line 179: `nil` is the default value for a function, so it shouldn't be added.

gnvk (commented):
Please extend or add a new unit test to test the new functionality!

marketdata/stream/client_test.go
```diff
@@ -207,6 +207,52 @@ func TestConnectSucceeds(t *testing.T) {
...
+
+			err := c.Connect(ctx)
+			require.NoError(t, err)
+			assert.Equal(t, 1, numConnectCalls)
+			assert.Equal(t, 0, numDisconnectCalls)
+
+			// Now force the stream to disconnect via context and assert disconnect is called after
+			// waiting a small amount of time to wait for the stream to shut down
+			cancel()
+			time.Sleep(100 * time.Millisecond)
```
gnvk, line 249: Please do not use sleep in unit tests. Here's how you should do this instead: in the disconnect callback function you should write to a channel, and here you should read from that channel with a timeout:
``` go
select {
case <-disconnects:
case <-time.After(time.Second):
	require.Fail(t, "disconnect callback was not called in time")
}
```

gnvk (approved):
LGTM. Thank you for this addition.

kalmant (commented):
General question: What about running these in their own goroutines?
That would simplify the code a lot and I think it would still fit the use cases described in the PR.

Edit: they are called in goroutines. But then why have those waitgroups?

marketdata/stream/options.go
```diff
@@ -102,6 +104,26 @@ func WithReconnectSettings(limit int, delay time.Duration) Option {
 	})
 }
 
+// WithConnectCallback runs the callback function after the streaming connection is setup.
+// If the stream terminates and can't reconnect, the disconnect callback will timeout one second
```
kalmant, line 108: > the disconnect callback will timeout
I think you meant to say "connect" here

marketdata/stream/client.go
```diff
@@ -440,10 +459,34 @@ func (c *client) maintainConnection(ctx context.Context, u url.URL, initialResul
...
+					c.disconnectCallback()
+				}()
+			}
 		}
 	}
 }
 
+// waitTimeout waits for the WaitGroup for the specified max timeout.
+// Returns true if waiting timed out.
+func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
```
kalmant, line 476: [minor]: I would perhaps rename this to `waitWithTimeout` and return false when there's a timeout instead of the other way around.
Both of these are minor comments so feel free to ignore them if you think the current solution is better!

kalmant (approved):
All my comments are minor comments so I'm approving. Thanks for this 👍🏼

Issue #211, opened 2022-09-20
The order entity does not match the full spec according to the docs


probably-not:
The current order entity does not match the full spec according to the docs, and is missing some fields

PR #212, opened 2022-09-20
Ensure order entity matches the full order entity spec

probably-not:


probably-not:
This will close #211

probably-not:
I haven't checked any of the other entities (I'm working mostly with the Orders entity right now) but I can do a quick review to make sure those match according to the spec as well

alpaca/entities.go
```diff
@@ -52,12 +52,13 @@ type Order struct {
...
 	AssetID        string           `json:"asset_id"`
 	Symbol         string           `json:"symbol"`
-	Exchange       string           `json:"exchange"`
 	Class          string           `json:"asset_class"`
+	OrderClass     OrderClass       `json:"order_class"`
 	Qty            *decimal.Decimal `json:"qty"`
 	Notional       *decimal.Decimal `json:"notional"`
 	FilledQty      decimal.Decimal  `json:"filled_qty"`
-	Type           OrderType        `json:"order_type"`
+	OrderType      OrderType        `json:"order_type"`
```
gnvk, line 60: `order_type` is deprecated, so it shouldn't be added, you can simply use `type` instead.

gnvk (approved):
LGTM. Thank you for your contribution.

PR #216, opened 2022-10-28
Add currency

gnvk:


marketdata/rest.go
```diff
@@ -158,22 +163,33 @@ func defaultDo(c *client, req *http.Request) (*http.Response, error) {
...
-func setBaseQuery(q url.Values, start, end time.Time, feed, defaultFeed, asof string) {
-	if !start.IsZero() {
-		q.Set("start", start.Format(time.RFC3339))
+type baseParams struct {
+	Start    time.Time
+	End      time.Time
+	Feed     string
+	AsOf     string
+	Currency string
+}
```
ccnlui, line 172: Do the members have to be exported when `baseParams` is not?

ccnlui (approved):
LGTM

marketdata/rest.go
gnvk: It doesn't really matter.

PR #219, opened 2022-11-17
Add side field to ListOrdersRequest

gfelbing:
According to the [API Doc](https://alpaca.markets/docs/api-references/trading-api/orders/#get-a-list-of-orders), there is a query parameter for `side` in the list orders endpoint.

This PR adds the field to the according entity.


gfelbing:
Fixes #220

gnvk (approved):
LGTM. Thank you for your contribution!

PR #226, opened 2023-01-17
Return order(s) from the close position methods

### LINKED ISSUE #124: Return order for ClosePosition()
The ClosePosition method create a market sell order for the symbol. Currently it only return err, is it ok to also return the sell order as well?

gnvk:
Fixes #124


said-saifi (approved):
LGTM

PR #236, opened 2023-01-24
Retry slightly after the last message so that it is not duplicated

gnvk:


ekovacs (approved):
since there is no input parameter at the server side to handle the param inclusive/exclusive, this is the best hack to avoid duplicate

PR #242, opened 2023-02-15
Retry on internal server errors

gnvk:


marketdata/rest.go
```diff
@@ -97,13 +97,17 @@ func defaultDo(c *Client, req *http.Request) (*http.Response, error) {
...
 		resp, err = c.httpClient.Do(req)
 		if err != nil {
 			return nil, err
 		}
-		if resp.StatusCode != http.StatusTooManyRequests {
-			break
+		switch resp.StatusCode {
+		case http.StatusTooManyRequests, http.StatusInternalServerError:
+		default:
+			break RetryLoop
```
Neal, line 110: doesn't this mean it will now always retry up to the RetryLimit even on successes?

gnvk: No. On success the status code is not 429 or 500, so we will break out of the loop on this line.

Neal: aah you're right. I misread and thought it'd continue here for some reason.

Neal (approved):

PR #245, opened 2023-03-10
PLTF-189: support new event streaming

turip:
Notable change is that for trade events the ID is now an ULID.

This means that when querying the since/until ID needs to be specified as an ULID and in the response the integer field is replaced by an ULID.

The new event streaming engine should be faster and more responsive event at market open than the previous one.


alpaca/entities.go
```diff
@@ -293,6 +294,7 @@ const (
 type TradeUpdate struct {
 	At          time.Time        `json:"at"`
 	Event       string           `json:"event"`
+	EventID     ulid.ULID        `json:"event_id"`
```
gnvk, line 297: This is more like a question: what's the benefit of using this type over string from the client's perspective? I don't think it would ever be used to anything other than copy/pasting to sinceID/untilID queries.

gnvk: We're using strings (instead of UUIDs) for all the other ID types in the codebase.

turip: Right, given that we support setting the since/until time, it won't give us much. The end user can still parse the ulid if needed. Let me make the necessary chanages.

turip: Fixed, please re-check the PR.

gnvk (approved):
LGTM. Thank you for adding these.

PR #246, opened 2023-03-13
TRD-1016: Increase reconnection limit to 40

AlexandrosKyriakakis:
### Context
Getting the following error frequently -> `max reconnect limit has been reached`


AlexandrosKyriakakis:
Addressed.

PR #252, opened 2023-07-27
Add AssetAttributes on assets entity

AlexandrosKyriakakis:


alpaca/entities.go
```diff
@@ -116,19 +116,28 @@ type Position struct {
...
+	Tradable                     bool             `json:"tradable"`
+	Marginable                   bool             `json:"marginable"`
+	MaintenanceMarginRequirement uint             `json:"maintenance_margin_requirement"`
+	Shortable                    bool             `json:"shortable"`
+	EasyToBorrow                 bool             `json:"easy_to_borrow"`
+	Fractionable                 bool             `json:"fractionable"`
+	Attributes                   []AssetAttribute `json:"attributes"`
 }
 
+type AssetAttribute string
```
gnvk, line 134: Any advantage of using a custom type here vs simple strings? Especially that this seems to be a highly flexible field.

AlexandrosKyriakakis: I used custom type just to be able to match the existing attributes `ptp_with_exception` `ptp_no_exception`.
I can update it with  `[]string` if you think it is better for this use case.

gnvk: You can easily match with strings. I'd just go with them.

gnvk (approved):
LGTM

PR #253, opened 2023-08-01
Replace decimal with alpacadecimal & bump go to 1.20

AlexandrosKyriakakis:
## Context
- Replacing `decimal` library with `alpacadecimal` -> https://github.com/alpacahq/alpacadecimal **Backwards Incompatible** 
- Bump go to 1.20


raja (approved):

gnvk (changes requested):
This is a backward incompatible change, which means that we would need to increase the major version to v4. I really don't think that this performance optimisation would be worth it. Is there a particular issue with the `decimal` packages performance you're facing?

gnvk:
This is a backward incompatible change. @AlexandrosKyriakakis I changed the base branch to v4 to reflect this. In itself it's not a big enough change to justify a v4 release, but we can keep it on the v4 branch and merge it to master later, possibly with bigger, more significant changes in the future.

PR #254, opened 2023-08-02
Add User-Agent header on HTTP requests and WSS

AlexandrosKyriakakis:
## Context 
- Adding User-Agent header on HTTP requests and WSS
- Similar to [alpaca-py](https://github.com/alpacahq/alpaca-py/blob/695f23de44c5410849b88acb77bdad09c327fb9f/alpaca/common/websocket.py#L78)

### DIFF
--- a/alpaca/version.go
+++ b/alpaca/version.go
@@ -1,3 +1,34 @@
 package alpaca
 
-const Version = "v3.1.1"
+import (
+	"runtime/debug"
+	"strings"
+	"sync"
+)
+
+const repoName = "github.com/alpacahq/alpaca-trade-api-go"
+
+var (
+	goVersion     string //
+	moduleVersion string
+	once          = sync.Once{}
+)
+
+// GetVersion returns running go version and alpaca-trade-api-go version
+func GetVersion() (string, string) {
+	once.Do(func() {
+		buildInfo, found := debug.ReadBuildInfo()
+		if !found {
+			return
+		}
+		goVersion = buildInfo.GoVersion
+
+		for _, dep := range buildInfo.Deps {
+			if strings.HasPrefix(dep.Path, repoName) {
+				moduleVersion = dep.Version
+				return
+			}
+		}
+	})
+	return goVersion, moduleVersion
+}
--- a/alpaca/version.go
+++ b/alpaca/version.go
@@ -1,3 +1,3 @@
 package alpaca
 
-const Version = "v3.1.0"
+const Version = "v3.1.1"
--- a/alpaca/version.go
+++ b/alpaca/version.go
@@ -4,30 +4,28 @@ import (
 	"runtime"
 	"runtime/debug"
 	"strings"
-	"sync"
 )
 
 const repoName = "github.com/alpacahq/alpaca-trade-api-go"
 
-var (
-	once            = sync.Once{}
-	encodedVersions string
-)
+var version = initVersion()
 
-// GetVersion returns running go version and alpaca-trade-api-go version
-func GetVersion() string {
-	once.Do(func() {
-		buildInfo, found := debug.ReadBuildInfo()
-		if !found {
-			return
+func initVersion() string {
+	var moduleVersion string
+	buildInfo, found := debug.ReadBuildInfo()
+	if !found {
+		return "GoRuntime/" + runtime.Version()
+	}
+	for _, dep := range buildInfo.Deps {
+		if strings.HasPrefix(dep.Path, repoName) {
+			moduleVersion += "APCA-GO/" + dep.Version + " "
+			break
 		}
-		for _, dep := range buildInfo.Deps {
-			if strings.HasPrefix(dep.Path, repoName) {
-				encodedVersions += "APCA-GO/" + dep.Version + " "
-				break
-			}
-		}
-		encodedVersions += "GoRuntime/" + runtime.Version()
-	})
-	return encodedVersions
+	}
+	return moduleVersion + "GoRuntime/" + runtime.Version()
+}
+
+// Version returns a string contains alpaca-trade-api-go dep version and go runtime version
+func Version() string {
+	return version
 }
--- a/alpaca/version.go
+++ b/alpaca/version.go
@@ -9,26 +9,30 @@ import (
 const repoName = "github.com/alpacahq/alpaca-trade-api-go"
 
 var (
-	goVersion     string
-	moduleVersion string
-	once          = sync.Once{}
+	once            = sync.Once{}
+	encodedVersions string
 )
 
 // GetVersion returns running go version and alpaca-trade-api-go version
-func GetVersion() (string, string) {
+func GetVersion() string {
 	once.Do(func() {
+		var (
+			goVersion     string
+			moduleVersion string
+		)
+
 		buildInfo, found := debug.ReadBuildInfo()
 		if !found {
 			return
 		}
 		goVersion = buildInfo.GoVersion
-
 		for _, dep := range buildInfo.Deps {
 			if strings.HasPrefix(dep.Path, repoName) {
 				moduleVersion = dep.Version
-				return
+				break
 			}
 		}
+		encodedVersions = "APCA-GO/" + moduleVersion + "/" + goVersion
 	})
-	return goVersion, moduleVersion
+	return encodedVersions
 }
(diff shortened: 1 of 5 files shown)


alpaca/version.go
```diff
@@ -0,0 +1,3 @@
+package alpaca
+
+const Version = "v3.1.1"
```
gnvk, line 3: Please consider using https://pkg.go.dev/runtime/debug#ReadBuildInfo instead.

raja: ^ex: ```go
build, ok := debug.ReadBuildInfo()
if !ok {
  log.Printf("Failed to read build info")
  return
}

for _, dep := range build.Deps {
  fmt.Printf("%s %s\n", dep.Path, dep.Version)
}
```

File: marketdata/rest.go [hunk 2]
```diff
@@ -88,6 +88,9 @@ func NewClient(opts ClientOpts) *Client {
 var DefaultClient = NewClient(ClientOpts{})
 
 func defaultDo(c *Client, req *http.Request) (*http.Response, error) {
+	goVer, moduleVer := alpaca.GetVersion()
+	req.Header.Set("User-Agent", "APCA-GO/"+moduleVer+"/"+goVer)
```
raja, line 92: why not encode formatted str in version.go?

raja (changes requested):

marketdata/stream/conn_nhoory.go
```diff
@@ -19,13 +21,15 @@ type nhooyrWebsocketConn struct {
 func newNhooyrWebsocketConn(ctx context.Context, u url.URL) (conn, error) {
 	ctxWithTimeout, cancel := context.WithTimeout(ctx, 3*time.Second)
 	defer cancel()
-
+	goVer, moduleVer := alpaca.GetVersion()
 	conn, _, err := websocket.Dial(ctxWithTimeout, u.String(), &websocket.DialOptions{
 		CompressionMode: websocket.CompressionContextTakeover,
 		HTTPHeader: http.Header{
 			"Content-Type": []string{"application/msgpack"},
+			"User-Agent":   []string{"APCA-GO/" + moduleVer + "/" + goVer},
```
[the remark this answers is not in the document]
AlexandrosKyriakakis: Nice Catch! 
Addressed: https://github.com/alpacahq/alpaca-trade-api-go/pull/254/commits/0a55aca0ed5eac63f4259b7f71116c04385650b8

marketdata/rest.go
AlexandrosKyriakakis: Addressed: https://github.com/alpacahq/alpaca-trade-api-go/pull/254/commits/0a55aca0ed5eac63f4259b7f71116c04385650b8

alpaca/version.go
```diff
@@ -0,0 +1,38 @@
...
+			return
+		}
+		goVersion = buildInfo.GoVersion
+		for _, dep := range buildInfo.Deps {
+			if strings.HasPrefix(dep.Path, repoName) {
+				moduleVersion = dep.Version
+				break
+			}
+		}
+		encodedVersions = "APCA-GO/" + moduleVersion + "/" + goVersion
```
leki75, line 35: minor: I suggest following https://www.rfc-editor.org/rfc/rfc9110.html#name-user-agent (or https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent).

Example:
`"ApcaGo/" + moduleVersion + " GoRuntime/" + goVersion`

alpaca/version.go
```diff
@@ -0,0 +1,38 @@
...
+		var (
+			goVersion     string
+			moduleVersion string
+		)
+
+		buildInfo, found := debug.ReadBuildInfo()
+		if !found {
+			return
+		}
+		goVersion = buildInfo.GoVersion
```
leki75, line 28: minor: I would stick to `runtime.Version()` instead of the build version, I think it is more helpful. Moreover, if build info is not available we can still return the header!

alpaca/version.go
```diff
@@ -0,0 +1,38 @@
...
+const repoName = "github.com/alpacahq/alpaca-trade-api-go"
+
+var (
+	once            = sync.Once{}
+	encodedVersions string
+)
+
+// GetVersion returns running go version and alpaca-trade-api-go version
+func GetVersion() string {
+	once.Do(func() {
```
leki75, line 18: minor: `init` is evil, but instead of `sync.Once` we can do it as https://github.com/uber-go/guide/blob/master/style.md#avoid-init suggests:
```go
const repo = "github.com/alpacahq/alpaca-trade-api-go"

var version = initVersion()

func initVersion() string {
    moduleVersion := "unknown"
    if buildInfo, ok := debug.ReadBuildInfo(); ok {
		for _, dep := range buildInfo.Deps {
			if strings.HasPrefix(dep.Path, repo) {
				moduleVersion = dep.Version
				break
			}
		}
    }
    return fmt.Sprintf("ApcaGo/%s GoRuntime/%s", moduleVersion, runtime.Version())
}

func Version() string {
    return version
}
```

marketdata/stream/conn_nhoory.go
```diff
@@ -19,13 +21,14 @@ type nhooyrWebsocketConn struct {
 func newNhooyrWebsocketConn(ctx context.Context, u url.URL) (conn, error) {
 	ctxWithTimeout, cancel := context.WithTimeout(ctx, 3*time.Second)
 	defer cancel()
-
 	conn, _, err := websocket.Dial(ctxWithTimeout, u.String(), &websocket.DialOptions{
 		CompressionMode: websocket.CompressionContextTakeover,
 		HTTPHeader: http.Header{
 			"Content-Type": []string{"application/msgpack"},
+			"User-Agent":   []string{alpaca.GetVersion()},
```
leki75, line 28: minor: While manipulating directly the underlying map is ok now, using the Set or Add methods of `http.Header` has the benefit of canonicalizing the header key.
> Get assumes that all keys are stored in canonical form. To use non-canonical keys, access the map directly.

leki75 (commented):
LGTM with minor comments

alpaca/version.go
raja, line 35:
```suggestion
		encodedVersions =  fmt."APCA-GO/" + moduleVersion + " GoRuntime/" + goVersion
```

now we are really being nitpicky :)

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent#syntax

ex:
```
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko)
Chrome/51.0.2704.103 Safari/537.36
Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:47.0) Gecko/20100101 Firefox/47.0
Mozilla/5.0 (platform; rv:geckoversion) Gecko/geckotrail Firefox/firefoxversion
```

raja:
@AlexandrosKyriakakis going to address reviews?

alpaca/version.go
AlexandrosKyriakakis: Addressed: https://github.com/alpacahq/alpaca-trade-api-go/pull/254/commits/783440c543b38222ed2b864afaa78626d5f9968e

alpaca/version.go
AlexandrosKyriakakis: Nice Catch! 🙏🏼 

Addressed: https://github.com/alpacahq/alpaca-trade-api-go/pull/254/commits/689e33c0262c4cc4e0f01fe704c8a2a8ea9248a8

marketdata/stream/conn_nhoory.go
AlexandrosKyriakakis: Addressed: https://github.com/alpacahq/alpaca-trade-api-go/pull/254/commits/858800b73d759b9614842be2439ff1c9e00a3407

Is this what we want?

alpaca/version.go
AlexandrosKyriakakis: Addressed: https://github.com/alpacahq/alpaca-trade-api-go/commit/783440c543b38222ed2b864afaa78626d5f9968e

leki75 (approved):
LGTM

AlexandrosKyriakakis:
Hey @raja I think I need also a re-review from you before merging.

raja (approved):

PR #255, opened 2023-08-17
Add sort parameter and re-add historical crypto quotes

gnvk:


leki75 (approved):
[minor]: I would separate this PR into 2:
* add sort parameter
* re-add historical crypto quotes

PR #259, opened 2023-09-04
Add Async and Paginated Functions (#1)

probably-not:
* Add PageToken to the requests

* Set page_token on base request

* Set page_token on base crypto request

* Add PageToken to NewsRequest too

* Add Paginated versions of each function that has a NextPageToken

* Add Paginated versions of each function that has a NextPageToken

* Fix returns

* Add Async Callback versions of each function that has a NextPageToken

* Consolidate

* Consolidate

* Replace page token on each iteration in async

* Total Limit to page limit or else max

### DIFF
--- a/marketdata/rest.go
+++ b/marketdata/rest.go
@@ -135,6 +135,11 @@ type baseRequest struct {
 	Sort     Sort
 }
 
+type basePaginatedRequest struct {
+	baseRequest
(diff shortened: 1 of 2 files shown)


probably-not:
My mistake, I opened a new PR instead of updating the existing one 🤦‍♂️

PR #268, opened 2023-12-06
Extend the calendar day with session and settlment info

### LINKED ISSUE #266: Additions to Calendar struct
Would it be possible to add the session open/close and settlement date to CalendarDay https://github.com/alpacahq/alpaca-trade-api-go/blob/efb7598b00a215d6ae06ca43b8592e7cbcb8e5c1/alpaca/entities.go#L177 

```
{
      "date": "1970-01-02",
      "open": "09:30",
      "close": "16:00",
      "session_open": "0400",
      "session_close": "2000",
      "settlement_date": "1970-01-06"
}
```

gnvk:
Fixes #266


alpaca/entities.go
```diff
@@ -175,9 +175,12 @@ type Fundamental struct {
...
 
 type CalendarDay struct {
-	Date  string `json:"date"`
-	Open  string `json:"open"`
-	Close string `json:"close"`
+	Date           string `json:"date"`
+	Open           string `json:"open"`
+	Close          string `json:"close"`
+	SessionOpen    string `json:"session_open"`
+	SessionClose   string `json:"session_close"`
```
Neal, line 182: I wouldn't like to add the session open/close fields here as they were exposed to the API unintentionally and there is no good use for them either as they don't really mean much as far as trading goes.

We can't remove it from the API response as that would be a breaking change, but at least we avoid adding it here to avoid more breaking changes in future.

gnvk: Can you please add this comment to the original issue: https://github.com/alpacahq/alpaca-trade-api-go/issues/266?

gnvk:
> @gnvk can we close this PR?

There's a kind of stuck discussion about this under this issue: https://github.com/alpacahq/alpaca-trade-api-go/issues/266 It would be nice to find a solution there.

DrudTheTerrible:
> > @gnvk can we close this PR?
> 
> There's a kind of stuck discussion about this under this issue: #266 It would be nice to find a solution there.

Issue #266 hasn't been updated in over a year. I think it's fine to close both the issue and PR if there is no intention to resolve it in short order.

PR #271, opened 2024-03-08
Add marketdata for options

gnvk:


examples/marketdata/marketdata.go
```diff
@@ -120,6 +122,53 @@ func cryptoQuote() {
...
+	if err != nil {
+		panic(err)
+	}
+	type snap struct {
+		marketdata.OptionSnapshot
+		Symbol string
+	}
+	calls, puts := []snap{}, []snap{}
+	for symbol, snapshot := range chain {
+		if strings.Contains(symbol, "C") {
```
leki75, line 136: This can be misleading if the user requests a symbol, that contains a letter 'C'. This is possibly too complex, but shows the idea:
```go
idx := len(symbol) - 9
if idx < 7 { // shortest symbol: A123456C12345678
    panic("invalid symbol")
}
switch (symbol[idx]) {
    case 'C':
        calls = append(calls, snap{OptionSnapshot: snapshot, Symbol: symbol})
    case 'P':
        puts = append(puts, snap{OptionSnapshot: snapshot, Symbol: symbol})
    default:
        panic("invalid symbol") 
}
```

marketdata/options.go
```diff
@@ -0,0 +1,346 @@
...
+		return nil, err
+	}
+	snapshot, ok := resp[symbol]
+	if !ok {
+		return nil, nil
+	}
+	return &snapshot, nil
+}
+
+// GetOptionSnapshots returns the snapshots for multiple symbol
```
leki75, line 250: typo: symbols

marketdata/stream/subscription.go
```diff
@@ -178,6 +178,40 @@ func (cc *CryptoClient) UnsubscribeFromOrderbooks(symbols ...string) error {
...
+	return cc.handleSubChange(false, subscriptions{bars: symbols})
+}
+
+func (cc *OptionClient) UnsubscribeFromUpdatedBars(symbols ...string) error {
+	return cc.handleSubChange(false, subscriptions{updatedBars: symbols})
+}
+
+func (cc *OptionClient) UnsubscribeFromDailyBars(symbols ...string) error {
+	return cc.handleSubChange(false, subscriptions{dailyBars: symbols})
+}
```
leki75, line 213: There are no subscribe methods for bars.

examples/marketdata/marketdata.go
gnvk: This is just an example. The user cannot "request a symbol", AAPL is hardcoded.

PR #274, opened 2024-03-26
fix: Remove sub changes request when timeout happens

hiohiohio:
Context:
- putting a request message into `c.subChanges` blocks process with having `c.pendingSubChangeMutex.Lock()` when subscribe request is called twice during disconnection
- this leads deadlock to wait `c.pendingSubChangeMutex.Lock()` in `connWriter()` when reconnecting

Changes:
- drain `c.subChanges` channel when timeout happens. expecting this happens only when disconnected as c.subChanges are consumed before the timeout by `connWriter()` if connected.

### DIFF
--- a/marketdata/stream/client_test.go
+++ b/marketdata/stream/client_test.go
@@ -659,18 +659,8 @@ func TestSubscriptionTwiceAcrossConnectionIssues(t *testing.T) {
 	// server accepts subscription
 	conn1.readCh <- serializeToMsgpack(t, []subWithT{
 		{
-			Type:         "subscription",
-			Trades:       trades1,
-			Quotes:       nil,
-			Bars:         nil,
-			UpdatedBars:  nil,
-			DailyBars:    nil,
-			Statuses:     nil,
-			LULDs:        nil,
-			CancelErrors: nil,
-			Corrections:  nil,
-			Orderbooks:   nil,
-			News:         nil,
+			Type:   "subscription",
+			Trades: trades1,
 		},
 	})
 	err = <-subRes
--- a/marketdata/stream/client_test.go
+++ b/marketdata/stream/client_test.go
@@ -3,6 +3,7 @@ package stream
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/url"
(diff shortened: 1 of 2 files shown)


marketdata/stream/client_test.go
```diff
@@ -600,6 +601,170 @@ func TestSubscriptionAcrossConnectionIssues(t *testing.T) {
...
+			Quotes:       nil,
+			Bars:         nil,
+			UpdatedBars:  nil,
+			DailyBars:    nil,
+			Statuses:     nil,
+			LULDs:        nil,
+			CancelErrors: nil,
+			Corrections:  nil,
+			Orderbooks:   nil,
+			News:         nil,
```
gnvk, line 673: [minor] No need for all these default values.

hiohiohio: Thank you! Removed https://github.com/alpacahq/alpaca-trade-api-go/pull/274/commits/df11ad7feb723e039ae8423e0cc4054e3f1fe538

marketdata/stream/client_test.go
```diff
@@ -600,6 +601,160 @@ func TestSubscriptionAcrossConnectionIssues(t *testing.T) {
...
+
+	mockTimeAfterCh <- time.Now()
+	err = <-subRes
+	assert.Error(t, err)
+	assert.True(t, errors.Is(err, ErrSubscriptionChangeTimeout), "actual: %s", err)
+
+	// establish 2nd connection
+	conn2 := newMockConn()
+	writeInitialFlowMessagesToConn(t, conn2, subscriptions{trades: trades1})
+	// writeInitialFlowMessagesToConn(t, conn2, subscriptions{})
```
gnvk, line 696: [minor] Please remove the commented code.

hiohiohio: Thank you! missed it. Removed https://github.com/alpacahq/alpaca-trade-api-go/pull/274/commits/80878d871ff2085a328485185fefd614277af525

gnvk (approved):
Thank you for fixing this @hiohiohio !

PR #284, opened 2024-06-25
Add missing AccountActivity and GetAccountActivitiesRequest fields

gcatlin:
GetAccountActivitiesRequest:
- PageToken
- Category

AccountActivity:
- OrderID and OrderStatus (for FILL activities only) [are documented](https://docs.alpaca.markets/reference/getaccountactivities-2)
- Status (for non-FILL activities) is not documented

### DIFF
--- a/alpaca/entities.go
+++ b/alpaca/entities.go
@@ -186,6 +186,9 @@ type AccountActivity struct {
 	NetAmount       decimal.Decimal `json:"net_amount"`
 	Description     string          `json:"description"`
 	PerShareAmount  decimal.Decimal `json:"per_share_amount"`
+	OrderID         string          `json:"order_id"`
+	OrderStatus     string          `json:"order_status"`
+	Status          string          `json:"status"`
 }
 
 //easyjson:json
--- a/alpaca/entities_easyjson.go
+++ b/alpaca/entities_easyjson.go
@@ -3017,6 +3017,12 @@ func easyjson3e8ab7adDecodeGithubComAlpacahqAlpacaTradeApiGoV3Alpaca23(in *jlexe
 			if data := in.Raw(); in.Ok() {
 				in.AddError((out.PerShareAmount).UnmarshalJSON(data))
 			}
+		case "order_id":
+			out.OrderID = string(in.String())
+		case "order_status":
+			out.OrderStatus = string(in.String())
+		case "status":
+			out.Status = string(in.String())
 		default:
 			in.SkipRecursive()
 		}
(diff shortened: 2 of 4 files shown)


alpaca/entities.go
```diff
@@ -186,6 +186,9 @@ type AccountActivity struct {
 	NetAmount       decimal.Decimal `json:"net_amount"`
 	Description     string          `json:"description"`
 	PerShareAmount  decimal.Decimal `json:"per_share_amount"`
+	OrderID         string          `json:"order_id"`
+	OrderStatus     string          `json:"order_status"`
+	Status          string          `json:"status"`
```
said-saifi, line 191: If you prefer, you can also add the `GroupID`, part of the non trade activity object and documented in [the docs](https://docs.alpaca.markets/reference/getaccountactivities-2).
Alternatively, I can create a subsequent PR to include this field.

said-saifi (approved):
Thanks, will update the docs.

alpaca/entities.go
gcatlin: I didn't add it because the response doesn't include the `group_id` field in the JSON. Can you verify that the server is returning it? I'll merge this one as is.

BTW, there are inconsistencies between the fields documented [here](https://docs.alpaca.markets/reference/getaccountactivities-2) and [here](https://docs.alpaca.markets/docs/account-activities).

On the second page the following fields are not documented:
- `order_status` (trade activity)
- `group_id` (non trade activity)

`status` (non trade activity) is missing from both pages.

said-saifi: The `group_id` field has an `omitempty` tag, meaning it will only appear in the response when populated. This field serves to link related activities together, primarily in options trading scenarios. For instance, when exercising an option, two activities are typically generated: "OPEXC" (Option Exercise) and "OPTRD" (Option Trade). These activities share the same group_id, indicating their association.

As for the inconsistency you stated, we are currently assessing if we really need the second page. If deemed unnecessary, users will be directed to refer to the primary page under the OpenAPI tab and we will delete the second page you linked. Otherwise, I'll update it.

Please note that for docs pages under the `Documentation` tab, you can suggest edits directly.

gcatlin: I don't use options, so adding `group_id` is low priority for me.

If you do delete the [second page](https://docs.alpaca.markets/docs/account-activities), please be sure to migrate the query parameter info from the "Pagination of Results" section to the Trader API docs  ([1](https://docs.alpaca.markets/reference/getaccountactivities-2), [2)](https://docs.alpaca.markets/reference/getaccountactivitiesbyactivitytype-1). The documentation in the Broker API docs ([1](https://docs.alpaca.markets/reference/getaccountactivities), [2](https://docs.alpaca.markets/reference/getaccountactivitiesbytype-1)) have this info. Also please migrate the "Activity Types" section that describes all the activity types (which will need to be updated because it doesn't match the API docs).

Incidentally, how does paginating through non-trade account activities work given that they only have a date field, but no higher resolution date/time field? I know there's page token, but how is activity order determined? I assume you have date/time field in your database. If so, could that be exposed by the API?

said-saifi: > Incidentally, how does paginating through non-trade account activities work given that they only have a date field, but no higher resolution date/time field? I know there's page token, but how is activity order determined? I assume you have date/time field in your database. If so, could that be exposed by the API?

Yes, our database includes a date-time column. However, since this column is not exposed in the API, sorting is performed using the id field. This id is a composite of system_date and a UUID, formatted as `20240626000000000::d80143f8-88db-4aaa-9a6b-c6c9a4d7c442`. This ensures that each ID is unique and sortable by the date component embedded within it.
Right now we don't see a need to expose the activity created at, but this might change in the future.

gnvk (approved):

PR #294, opened 2024-07-03
Add position intent to PlaceOrderRequest

### LINKED ISSUE #289: PlaceOrderRequest is missing `position_intent` query param
It would be good to have constants for the intents; e.g. `BuyToOpen`, `SellToClose`, etc.

https://docs.alpaca.markets/reference/postorder

ankitsridhar16:
Resolves #289 
- PlaceOrderRequest is missing position_intent query param

### DIFF
--- a/alpaca/entities.go
+++ b/alpaca/entities.go
@@ -205,6 +205,15 @@ type PortfolioHistory struct {
 
 type Side string
 
+type PositionIntent string
+
+const (
+	BuyToOpen   PositionIntent = "buy_to_open"
+	BuyToClose  PositionIntent = "buy_to_close"
+	SellToOpen  PositionIntent = "sell_to_open"
+	SellToClose PositionIntent = "sell_to_close"
+)
+
 const (
 	Buy  Side = "buy"
 	Sell Side = "sell"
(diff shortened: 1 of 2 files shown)


alpaca/entities.go
```diff
@@ -205,6 +205,15 @@ type PortfolioHistory struct {
...
+type PositionIntent string
+
+const (
+	BuyToOpen   PositionIntent = "buy_to_open"
+	BuyToClose  PositionIntent = "buy_to_close"
+	SellToOpen  PositionIntent = "sell_to_open"
+	SellToClose PositionIntent = "sell_to_close"
+)
+
 const (
```
gcatlin, line 217: Please move the `Side` constants to be just below their type declaration on line 206.

gcatlin (approved):
Other than the issue mentioned re: `Side`, this looks good to me. @gnvk might want you to update `rest_test.go`.

Thanks for doing this work!

alpaca/entities.go
ankitsridhar16: @gcatlin I have added the change, please verfiy

gcatlin: Looks good. Thanks!

PR #295, opened 2024-07-03
added missing fields to `Account` model

### LINKED ISSUE #288: Account model is missing fields returned by the API
https://docs.alpaca.markets/reference/getaccount-1

- `admin_configurations`
- `user_configurations`
- `options_approved_level`
- `options_trading_level`
- `options_buying_power`
- `pending_transfer_in`
- `balance_asof`
- `intraday_adjustments`
- `pending_reg_taf_fees`

ankitsridhar16:
resolves #288 
- Account model is missing fields returned by the API

### DIFF
--- a/alpaca/entities.go
+++ b/alpaca/entities.go
@@ -52,6 +52,12 @@ type Account struct {
 	SMA                   decimal.Decimal `json:"sma"`
 	DaytradeCount         int64           `json:"daytrade_count"`
 	CryptoTier            int             `json:"crypto_tier"`
+	OptionsTradingLevel   int             `json:"options_trading_level"`
+	PendingRegTafFees     string          `json:"pending_reg_taf_fees"`
+	OptionsApprovedLevel  int             `json:"options_approved_level"`
+	OptionsBuyingPower    string          `json:"options_buying_power"`
+	PendingTransferIn     string          `json:"pending_transfer_in"`
+	PendingTransferOut    string          `json:"pending_transfer_out"`
 }
 
 type Order struct {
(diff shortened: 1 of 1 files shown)


gcatlin (approved):
Looks good to me. Thanks for doing this!

Some **potential** changes, depending on @gnvk's preferences:
- change `BalanceAsof` to `BalanceAsOf` (capital "O") -- this reads better, but might be considered inconsistent with the JSON field name
- keep the `Options*` fields grouped together and the `Pending*` fields

ankitsridhar16:
> * Options*

I have added the suggested changes to the PR @gcatlin can you review ?

gcatlin:
@ankitsridhar16  Looks good. Thanks!

alpaca/entities.go
```diff
@@ -52,6 +52,14 @@ type Account struct {
...
 	DaytradeCount         int64           `json:"daytrade_count"`
 	CryptoTier            int             `json:"crypto_tier"`
+	OptionsTradingLevel   int             `json:"options_trading_level"`
+	OptionsApprovedLevel  int             `json:"options_approved_level"`
+	OptionsBuyingPower    string          `json:"options_buying_power"`
+	PendingRegTafFees     string          `json:"pending_reg_taf_fees"`
+	PendingTransferIn     string          `json:"pending_transfer_in"`
+	PendingTransferOut    string          `json:"pending_transfer_out"`
+	IntradayAdjustments   string          `json:"intraday_adjustments"`
+	BalanceAsOf           string          `json:"balance_asof"`
```
gnvk, line 62: This is a date, so I'd recommend to use `civil.Date` instead of a string.

gnvk (changes requested):
Please add / extend a unit test. If you do so, you will realise that the newly added fields are always empty, even when the server returns non-empty values for them. To fix that you'll need to regenerate the easyjson code with `make generate`.

ankitsridhar16:
> Please add / extend a unit test. If you do so, you will realise that the newly added fields are always empty, even when the server returns non-empty values for them. To fix that you'll need to regenerate the easyjson code with `make generate`.

I will check and ask if I have more doubts

PR #297, opened 2024-07-19
Add logging and optional callback to buffer fills

leki75:
Slow client related issues:
* `client.maintainConnection` waits for the `messageProcessor` goroutine(s) to finish to reconnect: this delays the reconnect and more messages can be missed
* whenever the client buffer `c.in` is full the `connReader` goroutine is blocked: the TCP stack of the operating system will buffer the messages and can cause disconnects if the application gets the pong message late, moreover the `connReader` goroutine will not exit if the context is cancelled while it is blocked on writing to `c.in`

### DIFF
--- a/marketdata/stream/client.go
+++ b/marketdata/stream/client.go
@@ -33,8 +33,9 @@ type client struct {
 	in                 chan []byte
 	subChanges         chan []byte
 
-	lastBufferFillMsg time.Time
-	droppedMsgs       int
+	bufferFillCallback func([]byte)
+	lastBufferFill     time.Time
+	droppedMsgCount    int
 
 	sub subscriptions
 
@@ -61,6 +62,7 @@ func (c *client) configure(o options) {
 	c.reconnectLimit = o.reconnectLimit
 	c.reconnectDelay = o.reconnectDelay
 	c.connectCallback = o.connectCallback
+	c.bufferFillCallback = o.bufferFillCallback
 	c.disconnectCallback = o.disconnectCallback
 	c.processorCount = o.processorCount
 	c.bufferSize = o.bufferSize
@@ -599,13 +601,16 @@ func (c *client) connReader(
 		select {
 		case c.in <- msg:
 		default:
-			c.droppedMsgs++
+			c.droppedMsgCount++
 			now := time.Now()
 			// Reduce the number of logs to 1 msg/sec if client buffer is full
-			if now.Add(-1 * time.Second).After(c.lastBufferFillMsg) {
-				c.logger.Warnf("datav2stream: writing to buffer failed, error: buffer full, dropped: %d", c.droppedMsgs)
-				c.droppedMsgs = 0
-				c.lastBufferFillMsg = now
+			if now.Add(-1 * time.Second).After(c.lastBufferFill) {
+				c.logger.Warnf("datav2stream: writing to buffer failed, error: buffer full, dropped: %d", c.droppedMsgCount)
+				c.droppedMsgCount = 0
+				c.lastBufferFill = now
+			}
+			if c.bufferFillCallback != nil {
+				c.bufferFillCallback(msg)
 			}
 		}
 	}
--- a/marketdata/stream/client.go
+++ b/marketdata/stream/client.go
@@ -33,6 +33,9 @@ type client struct {
 	in                 chan []byte
 	subChanges         chan []byte
 
+	lastBufferFillMsg time.Time
+	droppedMsgs       int
+
 	sub subscriptions
 
 	handler msgHandler
@@ -387,6 +390,17 @@ func (c *client) maintainConnection(ctx context.Context, u url.URL, initialResul
 		}
 	}
 
+	c.in = make(chan []byte, c.bufferSize)
+	pwg := sync.WaitGroup{}
+	pwg.Add(c.processorCount)
+	for i := 0; i < c.processorCount; i++ {
+		go c.messageProcessor(ctx, &pwg)
+	}
+	defer func() {
+		close(c.in)
+		pwg.Wait()
+	}()
+
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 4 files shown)


marketdata/stream/client.go
```diff
@@ -586,7 +596,18 @@ func (c *client) connReader(
...
+		case c.in <- msg:
+		default:
+			c.droppedMsgs++
+			now := time.Now()
+			// Reduce the number of logs to 1 msg/sec if client buffer is full
+			if now.Add(-1 * time.Second).After(c.lastBufferFillMsg) {
+				c.logger.Warnf("datav2stream: writing to buffer failed, error: buffer full, dropped: %d", c.droppedMsgs)
+				c.droppedMsgs = 0
+				c.lastBufferFillMsg = now
+			}
```
leki75, line 609: We should possibly call a handler function instead.

alanvoss (commented):
Other than my hope that someday we can become diligent about having tests for most code changes, and a couple of naming nits, this looks good to me.

marketdata/stream/client.go
```diff
@@ -33,6 +33,9 @@ type client struct {
 	in                 chan []byte
 	subChanges         chan []byte
 
+	lastBufferFillMsg time.Time
+	droppedMsgs       int
```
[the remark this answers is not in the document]
gnvk: Please do add a unit test for this change. `client_test.go` has some similar tests.

I'd prefer `lastBufferFillMsg` -> `lastBufferFill`, because it's not a timeout, it's the last time it was filled. I slightly prefer `droppedMsgs` over `droppedMsgsCount` or `droppedMsgCount` but either is fine.

marketdata/stream/client.go
gnvk: Yes, it would be nice. If you do that, please create an option (e.g. `WithBufferFillCallback`), similar to [WithDisconnectCallback](https://github.com/alpacahq/alpaca-trade-api-go/blob/8642da98f386b699a9abca62935630e77e4775e2/marketdata/stream/options.go#L130).

marketdata/stream/client_test.go
```diff
@@ -841,6 +842,90 @@ func TestSubscribeFailsDueToError(t *testing.T) {
...
+
+	sumTrades := 0
+	sumFills := 0
+	for i := 0; i < minFills; i++ {
+		select {
+		case trade := <-bufferFills:
+			sumFills++
+			assert.LessOrEqual(t, int64(minID), trade.ID)
+			assert.GreaterOrEqual(t, int64(maxID), trade.ID)
+		case <-timer.C:
```
gnvk, line 858: [minor] Replace with `<-time.After(100 * time.Millisecond)`?

marketdata/stream/options.go
```diff
@@ -123,6 +124,17 @@ func WithConnectCallback(callback func()) Option {
 	})
 }
 
+// WithBufferFillCallback runs the callback function whenever the buffer is full
+// and msg cannot be delivered. This usually happens when trade/quote handlers
+// process the messages slowly and it cannot keep up with the pace how messages
```
gnvk, line 129: [minor] **they** cannot keep up

marketdata/stream/client_test.go
```diff
@@ -841,6 +842,90 @@ func TestSubscribeFailsDueToError(t *testing.T) {
...
+	require.NoError(t, c.Connect(ctx))
+
+	// The buffer size is 2 but we send at least 4 (2 buffer size, 1
+	// messageProcessor goroutine, 1 extra) trades to have a buffer fill. The
+	// messageProcessor goroutines can read c.in while the rest of messages can
+	// be queued in the buffered channel.
+	connection.readCh <- serializeToMsgpack(t, []interface{}{tradeWithT{Type: "t", Symbol: "ALPACA", ID: 1}})
+	connection.readCh <- serializeToMsgpack(t, []interface{}{tradeWithT{Type: "t", Symbol: "ALPACA", ID: 2}})
+	connection.readCh <- serializeToMsgpack(t, []interface{}{tradeWithT{Type: "t", Symbol: "ALPACA", ID: 3}})
+	connection.readCh <- serializeToMsgpack(t, []interface{}{tradeWithT{Type: "t", Symbol: "ALPACA", ID: 4}})
```
gnvk, line 917: [minor] I'd use a for loop here. Same when sending trade 5..10.

marketdata/stream/client_test.go
```diff
@@ -841,6 +842,90 @@ func TestSubscribeFailsDueToError(t *testing.T) {
...
+		withConnCreator(func(ctx context.Context, u url.URL) (conn, error) { return connection, nil }),
+		WithTrades(func(t Trade) { trades <- t }, "ALPACA"),
+	)
+	require.NoError(t, c.Connect(ctx))
+
+	// The buffer size is 2 but we send at least 4 (2 buffer size, 1
+	// messageProcessor goroutine, 1 extra) trades to have a buffer fill. The
+	// messageProcessor goroutines can read c.in while the rest of messages can
+	// be queued in the buffered channel.
+	connection.readCh <- serializeToMsgpack(t, []interface{}{tradeWithT{Type: "t", Symbol: "ALPACA", ID: 1}})
```
gnvk, line 914: [minor] `interface{}` -> `any`

marketdata/stream/client_test.go
```diff
@@ -841,6 +842,90 @@ func TestSubscribeFailsDueToError(t *testing.T) {
...
+	bufferFills := make(chan Trade, 10)
+	trades := make(chan Trade)
+
+	c := NewStocksClient(marketdata.IEX,
+		WithBufferSize(bufferSize),
+		WithBufferFillCallback(func(msg []byte) {
+			trades := []tradeWithT{}
+			if err := msgpack.Unmarshal(msg, &trades); err != nil {
+				require.Fail(t, "msgpack unmarshal error")
+			}
```
gnvk, line 899:
```suggestion
			require.NoError(t, msgpack.Unmarshal(msg, &trades))
```

gnvk (approved):
Thank you @leki75 !

PR #310, opened 2024-11-11
add GetOptionContracts & GetOptionContract

nathanhack:
This adds GetOptionContracts and GetOptionContract support.


nathanhack:
I figure this is a good starting point for a discussion.

alpaca/options.go
```diff
@@ -0,0 +1,94 @@
+package alpaca
+
+import (
+	"cloud.google.com/go/civil"
+	"github.com/shopspring/decimal"
+)
+
+//go:generate go install github.com/mailru/easyjson/...@v0.7.7
+//go:generate easyjson -all -snake_case $GOFILE
```
gnvk, line 9: It's a bit strange that the "option entities" are now in `options.go` (not in `entities.go`), but the corresponding endpoint methods still in `rest.go`. I did it the other way around when adding options to the marketdata package: moving the entities to `entities.go` but having the methods in `options.go`. I'm fine with either of these solutions:
- no new file, add only to `entities.go` / `rest.go`
- all new options related stuff to `options.go`
- new options entities to `entities.go`, methods to `options.go`

alpaca/rest.go
```diff
@@ -742,6 +744,121 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+	ExpirationDate    *civil.Date      `json:"expiration_date"`
+	ExpirationDateGTE *civil.Date      `json:"expiration_date_gte"`
+	ExpirationDateLTE *civil.Date      `json:"expiration_date_lte"`
+	RootSymbol        *string          `json:"root_symbol"`
+	Type              *OptionType      `json:"type"`
+	Style             *OptionStyle     `json:"style"`
+	StrikePriceGTE    *decimal.Decimal `json:"strike_price_gte"`
+	StrikePriceLTE    *decimal.Decimal `json:"strike_price_lte"`
+	Limit             *int             `json:"limit"`
+	PPInd             *bool            `json:"ppind"`
```
gnvk, line 760: Recently I started not using pointers to optional fields for two main reasons:
1. It's much more comfortable to use:
``` go
showDeliverable := true
req := GetOptionContractsRequest{ShowDeliverable: &showDeliverable}
```
vs
``` go
GetOptionContractsRequest{ShowDeliverable: true}
```
2. There is no meaningful difference between a missing field vs a field with zero value.

alpaca/rest.go
```diff
@@ -742,6 +744,121 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
 	return &asset, nil
 }
 
+type GetOptionContractsRequest struct {
+	UnderlyingSymbols string           `json:"underlying_symbols"`
```
gnvk, line 748: Btw this is also optional, but see my other comment.

alpaca/rest.go
gnvk, line 760: I'd maybe spell out `PennyProgramIndicator` instead of `PPInd`

alpaca/rest.go
```diff
@@ -742,6 +744,121 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+			if subFieldType.Kind() == reflect.String {
+				v := subFieldValue.String()
+				if v != "" {
+					q.Set(fieldType.Tag.Get("json"), v)
+				}
+			} else {
+				q.Set(fieldType.Tag.Get("json"), fmt.Sprint(subFieldValue))
+			}
+		}
+	}
```
gnvk, line 809: Not sure if this reflection-based solution would work with my proposed zero-value instead of pointer approach. However, honestly I prefer writing "empty check" for every single field.

alpaca/rest.go
```diff
@@ -742,6 +744,121 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+	Status            *OptionStatus    `json:"status"`
+	ExpirationDate    *civil.Date      `json:"expiration_date"`
+	ExpirationDateGTE *civil.Date      `json:"expiration_date_gte"`
+	ExpirationDateLTE *civil.Date      `json:"expiration_date_lte"`
+	RootSymbol        *string          `json:"root_symbol"`
+	Type              *OptionType      `json:"type"`
+	Style             *OptionStyle     `json:"style"`
+	StrikePriceGTE    *decimal.Decimal `json:"strike_price_gte"`
+	StrikePriceLTE    *decimal.Decimal `json:"strike_price_lte"`
+	Limit             *int             `json:"limit"`
```
gnvk, line 759: Please provide `PageLimit` and `TotalLimit` options! See for example `GetOptionTradesRequest` for reference.

alpaca/rest.go
```diff
@@ -742,6 +744,121 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+	for {
+		u.RawQuery = q.Encode()
+
+		resp, err := c.get(u)
+		if err != nil {
+			return nil, err
+		}
+
+		// bs, err := ioutil.ReadAll(resp.Body)
+		// fmt.Println(string(bs))
```
gnvk, line 821: [minor] leftover

alpaca/rest.go
```diff
@@ -742,6 +744,121 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+			return nil, err
+		}
+
+		optionContracts = append(optionContracts, response.OptionContracts...)
+
+		if response.NextPageToken == nil {
+			break
+		}
+
+		fmt.Println("setting page_token:", *response.NextPageToken)
```
gnvk, line 834: [minor] leftover

alpaca/rest.go
```diff
@@ -742,6 +744,121 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+		q.Set("page_token", *response.NextPageToken)
+		closeResp(resp)
+	}
+
+	return optionContracts, nil
+}
+
+// GetOptionContract returns the Option Contracts.
```
gnvk, line 842:
```suggestion
// GetOptionContract returns an option contract by symbol or contract ID.
```

alpaca/rest_test.go
```diff
@@ -908,6 +909,123 @@ func TestGetAssetFromJSON(t *testing.T) {
...
+		assert.Equal(t, request.UnderlyingSymbols, req.URL.Query().Get("underlying_symbols"))
+		assert.Equal(t, strconv.FormatBool(*request.ShowDeliverable), req.URL.Query().Get("show_deliverables"))
+		assert.Equal(t, fmt.Sprint(*request.Status), req.URL.Query().Get("status"))
+		assert.Equal(t, fmt.Sprint(*request.ExpirationDate), req.URL.Query().Get("expiration_date"))
+		assert.Equal(t, *request.RootSymbol, req.URL.Query().Get("root_symbol"))
+		assert.Equal(t, fmt.Sprint(*request.Type), req.URL.Query().Get("type"))
+		assert.Equal(t, fmt.Sprint(*request.Style), req.URL.Query().Get("style"))
+		assert.Equal(t, fmt.Sprint(*request.StrikePriceGTE), req.URL.Query().Get("strike_price_gte"))
+		assert.Equal(t, strconv.Itoa(*request.Limit), req.URL.Query().Get("limit"))
+		assert.Equal(t, strconv.FormatBool(*request.PPInd), req.URL.Query().Get("ppind"))
```
gnvk, line 970: These asserts can be simplified a lot as well if we drop the pointers.

alpaca/options.go
nathanhack: I'll go with the no new file.

alpaca/rest.go
nathanhack: Oops, the changes from above will fix this.

nathanhack (commented):
Should be ready for another look.

alpaca/rest.go
```diff
@@ -742,6 +743,155 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+
+	q := u.Query()
+
+	if req.UnderlyingSymbols != "" {
+		q.Set("underlying_symbols", req.UnderlyingSymbols)
+	}
+
+	q.Set("show_deliverables", strconv.FormatBool(req.ShowDeliverable))
+
+	if req.Status == OptionStatusActive || req.Status == OptionStatusInactive {
```
gnvk, line 802: [minor] I'd slightly prefer `req.Status != ""`. That way if a new status is added, the user can use it even without updating the SDK.

alpaca/rest.go
```diff
@@ -742,6 +743,155 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+		q.Set("status", string(req.Status))
+	}
+
+	defaultDate := civil.Date{}
+	if req.ExpirationDate != defaultDate {
```
gnvk, line 807: [minor] I prefer `req.ExpirationDate.IsZero()`. Same for the other date checks.

alpaca/rest.go
gnvk: Same for the type and style check.

alpaca/rest.go
```diff
@@ -742,6 +743,155 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+	if req.Type == OptionTypeCall || req.Type == OptionTypePut {
+		q.Set("type", string(req.Type))
+	}
+
+	if req.Style == OptionStyleAmerican || req.Style == OptionStyleEuropean {
+		q.Set("style", string(req.Style))
+	}
+
+	defaultStrikePrice := decimal.Decimal{}
+	if req.StrikePriceLTE.Cmp(defaultStrikePrice) != 0 {
```
gnvk, line 832: [minor] `req.StrikePriceLTE.IsZero()`. Same for GTE.

alpaca/rest.go
```diff
@@ -742,6 +743,155 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+		q.Set("strike_price_lte", req.StrikePriceLTE.String())
+	}
+
+	if req.StrikePriceGTE.Cmp(defaultStrikePrice) != 0 {
+		q.Set("strike_price_gte", req.StrikePriceGTE.String())
+	}
+
+	q.Set("ppind", strconv.FormatBool(req.PennyProgramIndicator))
```
gnvk, line 840: This seems to be broken on the server side:
``` shell
$ curl -H "APCA-API-KEY-ID: ${APCA_API_KEY_ID}" -H "APCA-API-SECRET-KEY: ${APCA_API_SECRET_KEY}" \
    "https://paper-api.alpaca.markets/v2/options/contracts?ppind=false"
{"code":40010001,"message":"request parameters are invalid"}
```
I reported this issue internally, but for now, let's work around it by only setting this when it's true, like:
``` go
if req.PennyProgramIndicator {
	q.Set("ppind", "true")
}
```

alpaca/rest.go
```diff
@@ -742,6 +743,155 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+		}
+
+		var response optionContractsResponse
+		if err = unmarshal(resp, &response); err != nil {
+			return nil, err
+		}
+
+		optionContracts = append(optionContracts, response.OptionContracts...)
+
+		if response.NextPageToken == nil {
```
gnvk, line 864: You also need to break of the provided total limit is reached.

alpaca/rest.go
```diff
@@ -742,6 +743,155 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+}
+
+// GetOptionContracts returns the list of Option Contracts.
+func (c *Client) GetOptionContracts(req GetOptionContractsRequest) ([]OptionContract, error) {
+	u, err := url.Parse(fmt.Sprintf("%s/%s/options/contracts", c.opts.BaseURL, apiVersion))
+	if err != nil {
+		return nil, err
+	}
+
+	q := u.Query()
```
gnvk, line 794: One solution for the linter error would be to refactor building the query to a function.

nathanhack: But, if you want it in another function, you're the owner what ever you want goes.

nathanhack:
Should be good for another look.

alpaca/rest.go
gnvk: I actually prefer how it is now, as long as the linter is happy.

alpaca/rest_test.go
```diff
@@ -908,6 +909,116 @@ func TestGetAssetFromJSON(t *testing.T) {
 	assert.Nil(t, asset)
 }
 
+func TestGetOptionContracts(t *testing.T) {
+	c := DefaultClient
+	// successful with nils
```
gnvk, line 914: [minor] outdated comment, same for L939

alpaca/rest.go
```diff
@@ -742,6 +743,155 @@ func (c *Client) GetAsset(symbol string) (*Asset, error) {
...
+
+	if req.Type != "" {
+		q.Set("type", string(req.Type))
+	}
+
+	if req.Style != "" {
+		q.Set("style", string(req.Style))
+	}
+
+	if req.StrikePriceLTE.IsZero() {
```
gnvk, line 830: Your unit test actually caught this bug: should be negated. Same for `StrikePriceGTE`.

nathanhack: Sigh, I'm sad I missed this. But that's one of the reasons we write unit tests.

gnvk (approved):
LGTM. Thank you for your contribution @nathanhack !

PR #312, opened 2024-12-12
Add crypto perp endpoints

chathumi-k:


examples/crypto-perp-stream/crypto_stream.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/alpacahq/alpaca-trade-api-go/v3/marketdata/stream"
+)
+
+func main() {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	baseURL := "ws://stream.data.alpaca.markets/v1beta1/crypto-perps"
+
+	// Creating a client that connects to iex
```
gnvk, line 17: Not iex. But I think this comment is not useful anyway, so you can remove it completely.

examples/marketdata/marketdata.go
```diff
@@ -209,6 +236,9 @@ func main() {
 		{Name: "news", Func: news},
 		{Name: "auctions", Func: auctions},
 		{Name: "crypto_quote", Func: cryptoQuote},
+		{Name: "crypto_perp_quote", Func: cryptoPerpsQuote},
+		{Name: "crypto_perp_trade", Func: cryptoPerpsTrade},
+		{Name: "crypto_perp_bar", Func: cryptoPerpsBar},
```
gnvk, line 241: [minor] Could you please right-pad in line 248 now that we have more than 9 entries here?
``` go
fmt.Printf("[ %2d ] %s\n", i, e.Name)
```

marketdata/rest.go
```diff
@@ -44,6 +44,11 @@ type ClientOpts struct {
 	RequestHost string
 }
 
+type request interface {
```
gnvk, line 47: [minor] I'd call this `cryptoRequest`

marketdata/rest.go
```diff
@@ -766,9 +777,47 @@ func (c *Client) GetCryptoTrades(symbol string, req GetCryptoTradesRequest) ([]C
 	return resp[symbol], nil
 }
 
-// GetMultiTrades returns trades for the given crypto symbols.
+// GetCryptoPerpsTrades returns the trades for the given crypto perpetual futures symbol.
+func (c *Client) GetCryptoPerpsTrades(symbol string, req GetCryptoTradesRequest) ([]CryptoPerpsTrade, error) {
```
gnvk, line 781: `GetCryptoPerpsTrades` sounds strange to me, I'd prefer `GetCryptoPerpTrades`

marketdata/entities.go
```diff
@@ -208,6 +209,11 @@ type CryptoSnapshot struct {
 	PrevDailyBar *CryptoBar   `json:"prevDailyBar"`
 }
 
+type CryptoPerpsTrade CryptoTrade
```
gnvk, line 212: `CryptoPerpsTrade` -> `CryptoPerpTrade`. Same for all other types and methods, where "perp" is used as an adjective before a noun.

marketdata/rest.go
gnvk: [minor] Please move this to the top of the crypto section.

marketdata/rest.go
gnvk, line 781: Historical trades / quotes / bars are not supported by the server currently, so they should not be added.

marketdata/rest.go
```diff
@@ -983,6 +1084,26 @@ func (c *Client) cryptoFeed(fromReq string) string {
...
+			return fromReq.cryptoFeed()
+		}
+		if c.opts.CryptoFeed != "" {
+			return c.opts.CryptoFeed
+		}
+		if fromReq.isPerp() {
+			return "global"
+		}
+		return "us"
+	}(fromReq))
```
gnvk, line 1104: This feels a bit too functional for me in Go :P

```suggestion
	prefix := cryptoPrefix
	if fromReq.isPerp() {
		prefix = cryptoPerpPrefix
	}
	feed := fromReq.cryptoFeed()
	if feed == "" {
		if fromReq.isPerp() {
			feed = GLOBAL
		} else {
			feed = US
		}
	}
	return fmt.Sprintf("%s/%s/%s", c.opts.BaseURL, prefix, feed)
```

marketdata/rest.go
```diff
@@ -996,10 +1117,25 @@ func (c *Client) GetLatestCryptoBar(symbol string, req GetLatestCryptoBarRequest
 	return &bar, nil
 }
 
+// GetLatestCryptoPerpsBar returns the latest bar for a given crypto symbol
+func (c *Client) GetLatestCryptoPerpsBar(symbol string, req GetLatestCryptoBarRequest) (*CryptoPerpsBar, error) {
```
gnvk, line 1121: `GetLatestCryptoPerpsBars` for multiple symbols should also be added.

marketdata/rest.go
```diff
@@ -996,10 +1117,25 @@ func (c *Client) GetLatestCryptoBar(symbol string, req GetLatestCryptoBarRequest
 	return &bar, nil
 }
 
+// GetLatestCryptoPerpsBar returns the latest bar for a given crypto symbol
```
gnvk, line 1120: a given crypto perpetual future

marketdata/stream/client.go
```diff
@@ -216,6 +221,24 @@ func (cc *CryptoClient) constructURL() (url.URL, error) {
...
+	CryptoClient
+}
+
+// NewCryptoPerpsClient returns a new CryptoPerpsClient that will connect to the crypto perpetual futures feed
+// Currently only the marketdata.GLOBAL is supported.
+// Base URL will be modified to reflect the perpetual endpoint with respect to the crypto endpoint
+func NewCryptoPerpsClient(feed marketdata.CryptoFeed, opts ...CryptoOption) *CryptoPerpsClient {
+	if feed != marketdata.GLOBAL {
+		return nil
+	}
```
gnvk, line 235: Don't do this, it will at best cause a panic and at worst a silent error.

marketdata/stream/client.go
```diff
@@ -216,6 +221,24 @@ func (cc *CryptoClient) constructURL() (url.URL, error) {
 	return constructURL(cc.baseURL, cc.feed)
 }
 
+// This will create a crytpo perpetual futures subscriptions to the Alpaca market data services
+type CryptoPerpsClient struct {
```
gnvk, line 225: I don't see a benefit of adding this for now. The `CryptoClient` can be used instead. If we add a new message type to the perpetual future stream, then a new client will be needed, but also it will require additional changes.

chathumi-k: Yes, for now there is no benefit. But we added `CryptoPerpsClient` because we assumed that in the future it will be needed for the `futures-pricing-data`.

marketdata/rest.go
```diff
@@ -1754,39 +1663,29 @@ func GetCryptoSnapshots(symbols []string, req GetCryptoSnapshotRequest) (map[str
...
-	return DefaultClient.GetLatestCryptoPerpsQuotes(symbols, req)
+// GetLatestCryptoPerpQuote returns the latest quote for a given crypto perpetual future
+func GetLatestCryptoPerpQuote(symbol string, req GetLatestCryptoQuoteRequest) (*CryptoPerpQuote, error) {
+	return DefaultClient.GetLatestCryptoPerpQuote(symbol, req)
 }
 
-// GetCryptoPerpsSnapshot returns the snapshot for a given crypto symbol
-func GetCryptoPerpsSnapshot(symbol string, req GetCryptoSnapshotRequest) (*CryptoPerpsSnapshot, error) {
-	return DefaultClient.GetCryptoPerpsSnapshot(symbol, req)
+// GetLatestCryptoPerpBar returns the latest bar for a given crypto perpetual future
```
gnvk, line 1681: `GetLatestCryptoPerpBars` is missing, and please put both that and this function under `GetLatestCryptoPerpQuotes`

examples/marketdata/marketdata.go
chathumi-k: Done.

examples/crypto-perp-stream/crypto_stream.go
chathumi-k: Done.

marketdata/rest.go
chathumi-k: Done.

marketdata/entities.go
chathumi-k: Done.

marketdata/rest.go
chathumi-k: Removed.

marketdata/rest.go
chathumi-k: Added the suggested change.

marketdata/rest.go
chathumi-k: Done.

marketdata/rest.go
chathumi-k: Done.

marketdata/rest.go
```diff
@@ -1123,10 +1337,39 @@ func (c *Client) GetCryptoSnapshot(symbol string, req GetCryptoSnapshotRequest)
 	return &snapshot, nil
 }
 
+// GetCryptoPerpsSnapshot returns the snapshot for a given crypto perps symbol
+func (c *Client) GetCryptoPerpsSnapshot(symbol string, req GetCryptoSnapshotRequest) (*CryptoPerpsSnapshot, error) {
```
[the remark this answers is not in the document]
chathumi-k: Removed.

marketdata/rest.go
chathumi-k: Done.

gnvk (approved):
Thank you for adding these new endpoint ✅

PR #313, opened 2024-12-13
Add order imbalance stock messages

leki75:
Imbalance messages are sent during Limit-up Limit-down trading halts on the stock market.

### DIFF
--- a/marketdata/stream/subscription.go
+++ b/marketdata/stream/subscription.go
@@ -111,12 +111,6 @@ func (sc *StocksClient) UnsubscribeFromLULDs(symbols ...string) error {
 	return sc.handleSubChange(false, subscriptions{lulds: symbols})
 }
 
-func (sc *StocksClient) UnregisterImbalances() {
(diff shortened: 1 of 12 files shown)


marketdata/stream/subscription.go
```diff
@@ -96,10 +103,20 @@ func (sc *StocksClient) UnsubscribeFromStatuses(symbols ...string) error {
...
 
+func (sc *StocksClient) UnsubscribeFromImbalances(symbols ...string) error {
+	return sc.handleSubChange(false, subscriptions{imbalances: symbols})
+}
+
 func (sc *StocksClient) UnsubscribeFromLULDs(symbols ...string) error {
 	return sc.handleSubChange(false, subscriptions{lulds: symbols})
 }
 
+func (sc *StocksClient) UnregisterImbalances() {
```
gnvk, line 114: This is not needed.

gnvk (approved):
LGTM. Please only merge once the server side change is available.

PR #316, opened 2024-12-20
Tests for crypto perpetual market data

chathumi-k:
This PR include written tests to cover the crypto perpetual market data functionality. Tests cover both ws stream and REST.


gnvk (approved):
LGTM. I'm still not a fan of duplicating the stream client for crypto perpetual futures, currently the tests are basically duplicates, too.

PR #320, opened 2025-01-14
Fixed wrong API version in URLs and added APIVersion as client option.

cesschneider:


gnvk:
I don't really understand what you're trying to achieve here. Could you please share the issue you're facing?

cesschneider:
I just suggested a parameter to configure API version as it's currently hardcoded in the URLs.

gnvk:
My main problem with this is that there are multiple Alpaca endpoints with multiple api versions. For example the two endpoint groups you changed (`crypto` and `crypto-perp`) actually have different versions: v1beta3 and v1beta1, respectively. Setting a single `ApiVersion` in `ClientOpts` forces the users to create multiple accounts for each endpoint group, which feels like a workaround.

Moreover, I don't really see a use-case for changing the api version in the SDK. The SDK implements a client for one (typically the latest) api version for each endpoint, why would you want to change that (without changing the implementation, too)?

If you don't mind, for these reasons, I'm closing this PR.

PR #323, opened 2025-01-20
Improve the marketdata examples

gnvk:


examples/marketdata/marketdata.go
```diff
@@ -111,63 +101,45 @@ func auctions() {
...
-}
-
-func cryptoPerpPricing() {
-	pricing, err := marketdata.GetLatestCryptoPerpPricing("BTC-PERP", marketdata.GetLatestCryptoPerpPricingRequest{})
-	if err != nil {
-		panic(err)
-	}
-	fmt.Printf("Latest crypto perp pricing: %+v\n\n", pricing)
-	fmt.Println()
-}
```
leki75, line 152: NOTE: This block is removed.

gnvk: Good catch. Added back.

alanvoss (approved):

PR #324, opened 2025-01-21
Update go to v1.21 and cloud.google.com/go to v0.118.0

andrewwillette:
The `cloud.google.com/go v0.99.0` dependency is causing build failures.


gnvk (approved):
LGTM. I'm slightly worried about the go version bump, but honestly even v1.21 has already reached end of life, so I'm okay with it.

PR #326, opened 2025-02-05
Update the trade events endpoint to use stable version: v2beta1 => v2

ekovacs:


gnvk (approved):
LGTM, tested it on my account.

Issue #327, opened 2025-02-10
Not optional field in GO SDK


iamwhil:
When creating a `PlaceOrderRequest` the `PositionIntent` will default to nothing. Not a nil pointer or "none."

This means that creating a PlaceOrderRequest when trading stocks (not options) you still need to specify _something_ for the `PositionIntent`

# Example `PlaceOrderRequest` for GSK for 2 positions - sell side, market type.

```
SELL REQUEST: {Symbol:GSK Qty:2 Notional:<nil> Side:sell Type:market TimeInForce:day LimitPrice:<nil> ExtendedHours:false StopPrice:<nil> ClientOrderID: OrderClass: TakeProfit:<nil> StopLoss:<nil> TrailPrice:<nil> TrailPercent:<nil> PositionIntent:None}invalid position_intent specified (HTTP 422, Code 40010001)
panic: invalid position_intent specified (HTTP 422, Code 40010001)
```

Please add a valid `none` option const here and possibly set the default `PositionIntent` to `none`

https://github.com/alpacahq/alpaca-trade-api-go/blob/4ca59cf0a2f6aa816cb4c971ecff3502cd927e77/alpaca/entities.go#L212C1-L219C2

PR #328, opened 2025-02-12
Make PositionIntent optional in PlaceOrderRequest

sai-g:
The field `PositionIntent` was introduced in https://github.com/alpacahq/alpaca-trade-api-go/pull/294 as mandatory field, should've been optional 

Fixes https://github.com/alpacahq/alpaca-trade-api-go/issues/327

### DIFF
--- a/alpaca/rest.go
+++ b/alpaca/rest.go
@@ -558,7 +558,7 @@ type PlaceOrderRequest struct {
 	StopLoss       *StopLoss        `json:"stop_loss"`
 	TrailPrice     *decimal.Decimal `json:"trail_price"`
 	TrailPercent   *decimal.Decimal `json:"trail_percent"`
-	PositionIntent *PositionIntent  `json:"position_intent"`
+	PositionIntent PositionIntent   `json:"position_intent,omitempty"`
 	Legs           []Leg            `json:"legs"` // mleg order legs
 }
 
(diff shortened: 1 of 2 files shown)


alpaca/rest.go
```diff
@@ -558,7 +558,7 @@ type PlaceOrderRequest struct {
 	StopLoss       *StopLoss        `json:"stop_loss"`
 	TrailPrice     *decimal.Decimal `json:"trail_price"`
 	TrailPercent   *decimal.Decimal `json:"trail_percent"`
-	PositionIntent PositionIntent   `json:"position_intent"`
+	PositionIntent *PositionIntent  `json:"position_intent"`
```
gnvk, line 561: I don't like this change. For one it's backward incompatible, but more importantly, it's inconvenient to use. Generally with `string` parameters and here too it makes no sense to differentiate the missing and the empty value. So I strongly prefer `PositionIntent` over `*PositionIntent`, with "" indicating that it's missing from the request.

My proposal instead is adding "omitempty" to the json tag. Also, write a unit test to make sure this works as intended. I tested it manually and it seems so.

sai-g: Thanks for your suggestion! I've addressed this in https://github.com/alpacahq/alpaca-trade-api-go/pull/328/commits/354e993b9d89b040997b65b8dd69c767636bb428

gnvk (approved):
LGTM. Thank you for fixing this @sai-g!

iamwhil:
Thanks for hitting this up!

PR #330, opened 2025-03-15
Replace regex with strings.Contains in isHTTP4xx

SahibYar:
Ran the test on my local machine Apple M3 Pro, 36 GB Ram 

The string-based approach (is_http_4xx_string) is significantly faster than the regex-based approach (is_http_4xx_regex).
	•	Regex approach: ~0.1275 seconds (for 100,000 executions)
	•	String matching approach: ~0.0334 seconds (for 100,000 executions)

The string-matching method is about 3.8x faster than the regex approach.

### DIFF
--- a/marketdata/stream/client.go
+++ b/marketdata/stream/client.go
@@ -5,7 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
-	"regexp"
+	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


gnvk (approved):
This shouldn't matter too much because this check only runs once per connection creation and only if there's an error. Still, an improvement is an improvement ✅

PR #331, opened 2025-03-16
Fix: Ensure proper closure of HTTP response body to prevent resource leak

SahibYar:
This PR fixes a potential resource leak issue in  code. This oversight could lead to resource exhaustion over time. ([Reference](https://stackoverflow.com/a/33238755/2949645))

Changes:

- Explicitly added `resp.Body.Close()` in all relevant places where HTTP requests are made.
- Ensures that resources are properly released after the response is handled.
- Added few nil check to avoid nil potential dereference panics. 

## Testing. 
I executed all test cases in following files. 
```
marketdata/rest_test.go
marketdata/options_test.go

alpaca/rest_test.go
alpaca/stream_test.go
alpaca/trading_test.go
```

### DIFF
--- a/alpaca/entities.go
+++ b/alpaca/entities.go
@@ -362,6 +362,11 @@ type APIError struct {
 }
 
 func APIErrorFromResponse(resp *http.Response) error {
+	if resp == nil {
+		return fmt.Errorf("response is nil")
+	}
+	defer resp.Body.Close()
+
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return err
--- a/alpaca/rest.go
+++ b/alpaca/rest.go
@@ -1342,6 +1342,10 @@ func unmarshal(resp *http.Response, v easyjson.Unmarshaler) error {
 }
 
 func closeResp(resp *http.Response) {
+	if resp == nil || resp.Body == nil {
+		return // Avoids panic if resp is nil
+	}
+
 	// The underlying TCP connection can not be reused if the body is not fully read
 	_, _ = io.Copy(io.Discard, resp.Body)
 	resp.Body.Close()
--- a/marketdata/rest.go
+++ b/marketdata/rest.go
@@ -1769,24 +1769,33 @@ func (c *Client) get(u *url.URL) (*http.Response, error) {
 }
 
 func unmarshal(resp *http.Response, v easyjson.Unmarshaler) error {
+	if resp == nil || resp.Body == nil {
+		return fmt.Errorf("response or response body is nil")
+	}
+
 	var (
-		reader io.ReadCloser
+		reader = resp.Body
 		err    error
 	)
 	switch resp.Header.Get("Content-Encoding") {
 	case "gzip":
 		reader, err = gzip.NewReader(resp.Body)
 		if err != nil {
+			resp.Body.Close() // Prevents resource leak if gzip.NewReader fails
 			return err
 		}
 		defer reader.Close()
 	default:
-		reader = resp.Body
+		defer resp.Body.Close() //Close the original response body if not gzipped.
 	}
 	return easyjson.UnmarshalFromReader(reader, v)
 }
 
 func closeResp(resp *http.Response) {
+	if resp == nil || resp.Body == nil {
+		return // Avoids panic if resp is nil
+	}
+
 	// The underlying TCP connection can not be reused if the body is not fully read
 	_, _ = io.Copy(io.Discard, resp.Body)
 	resp.Body.Close()


alpaca/entities.go
```diff
@@ -362,6 +363,11 @@ type APIError struct {
 }
 
 func APIErrorFromResponse(resp *http.Response) error {
+	if resp == nil {
+		return errors.New("response is response")
```
gnvk, line 367: Incorrect error message, should be: "response is nil".

alpaca/entities.go
```diff
@@ -362,6 +363,11 @@ type APIError struct {
...
+	}
+	defer resp.Body.Close()
```
gnvk, line 369: If you check all references of the `APIErrorFromResponse()` function, `resp.Body.Close()` is called. E.g. in `alpaca/rest.go:1334`. Moving the call here makes sense, but then those calls should be removed.

alpaca/entities.go
```diff
@@ -362,6 +363,11 @@ type APIError struct {
 }
 
 func APIErrorFromResponse(resp *http.Response) error {
+	if resp == nil {
```
gnvk, line 366: This check doesn't make much sense, because if the `resp` is nil, we panic before reaching this. Moreover, according to the documentation of `http.Client.Do` in case of a nil error the response body (and thus the response itself) cannot be nil.

alpaca/rest.go
```diff
@@ -1342,6 +1342,10 @@ func unmarshal(resp *http.Response, v easyjson.Unmarshaler) error {
 }
 
 func closeResp(resp *http.Response) {
+	if resp == nil || resp.Body == nil {
```
gnvk, line 1345: Same comment: neither of these can be nil.

marketdata/rest.go
```diff
@@ -1769,24 +1769,33 @@ func (c *Client) get(u *url.URL) (*http.Response, error) {
...
 	var (
-		reader io.ReadCloser
+		reader = resp.Body
 		err    error
 	)
 	switch resp.Header.Get("Content-Encoding") {
 	case "gzip":
 		reader, err = gzip.NewReader(resp.Body)
 		if err != nil {
+			resp.Body.Close() // Prevents resource leak if gzip.NewReader fails
```
gnvk, line 1784: We should call `closeResp` every time in a defer before calling `unmarshal()`. Calling it here doesn't really make sense, because it won't cover the happy (non-nil) path. 

That being said, if you check all `unmarshal()` references, we actually don't call `closeResp()` before all of them. For example in `GetMultiTrades` I don't think it's called. Adding those missing calls instead would be great in this PR!

marketdata/rest.go
```diff
@@ -1769,24 +1769,33 @@ func (c *Client) get(u *url.URL) (*http.Response, error) {
...
 		defer reader.Close()
 	default:
-		reader = resp.Body
+		defer resp.Body.Close() // Close the original response body if not gzipped.
 	}
 	return easyjson.UnmarshalFromReader(reader, v)
 }
 
 func closeResp(resp *http.Response) {
+	if resp == nil || resp.Body == nil {
```
gnvk, line 1795: Same comment: neither of these can be nil in a nil error case.

alpaca/rest.go
SahibYar: eceived resp.Body == nil when called c.delete(u) in following function

CancelOrder
CancelAllOrders


in alpaca/rest.go

alpaca/rest.go
```diff
@@ -1329,20 +1311,23 @@ func (c *Client) delete(u *url.URL) (*http.Response, error) {
...
 
-func closeResp(resp *http.Response) {
-	// The underlying TCP connection can not be reused if the body is not fully read
+func CloseResp(resp *http.Response) {
+	if resp == nil || resp.Body == nil {
+		return
+	}
+	// The underlying TCP connection cannot be reused if the body is not fully read
 	_, _ = io.Copy(io.Discard, resp.Body)
 	resp.Body.Close()
```
Ale-Cas, line 1332: why not defer this call in the first line of the func?

SahibYar: 1st line of functions are nil checks
```
if resp == nil || resp.Body == nil {
    return
}
```
then we have only 1 line before `resp.Body.Close()`

Ale-Cas: It still looks cleaner and more readable to me to have `defer resp.Body.Close()` in the first line of a func called CloseResp, rather than the last one. It doesn't change much, since in practice it's hard that either the reader or writer of io.Copy are closed and that call will panic, but it's still less cognitive load imo to defer that .Close call as the first thing.

alpaca/rest.go
```diff
@@ -1329,20 +1311,23 @@ func (c *Client) delete(u *url.URL) (*http.Response, error) {
 	return c.do(c, req)
 }
 
-func verify(resp *http.Response) error {
+func Verify(resp *http.Response) error {
```
gnvk, line 1314: Please do not make these exported. I understand the intent (to reuse them in the `marketdata` package), but that isn't worth making these public. On the long run we can consider introducing an internal package and refactor the common code between `alpaca` and `marketdata`, but for now let's just keep the duplication please.

alpaca/rest.go
```diff
@@ -111,10 +111,11 @@ func defaultDo(c *Client, req *http.Request) (*http.Response, error) {
 		if i >= c.opts.RetryLimit {
 			break
 		}
+		CloseResp(resp) // Close body before retrying
```
gnvk, line 114: This is actually a good catch 👍

alpaca/rest.go
```diff
@@ -128,11 +129,10 @@ func (c *Client) GetAccount() (*Account, error) {
 		return nil, err
 	}
 
-	resp, err := c.get(u)
+	resp, err := c.get(u) //nolint:bodyclose // Linter Error
```
gnvk, line 132: [minor] Please use the following nolint (you can easily just replace all): 
``` go
resp, err := c.get(u) //nolint:bodyclose // unmarshal closes the body
```

marketdata/stream/conn_coder.go
```diff
@@ -24,11 +24,12 @@ func newCoderWebsocketConn(ctx context.Context, u url.URL) (conn, error) {
...
 	reqHeader.Set("Content-Type", "application/msgpack")
 	reqHeader.Set("User-Agent", alpaca.Version())
-	//nolint:bodyclose // According to its docs: you never need to close resp.Body yourself
-	conn, _, err := websocket.Dial(ctxWithTimeout, u.String(), &websocket.DialOptions{
+	//nolint:bodyclose // Linter Error
+	conn, ignoredResponse, err := websocket.Dial(ctxWithTimeout, u.String(), &websocket.DialOptions{
 		CompressionMode: websocket.CompressionContextTakeover,
 		HTTPHeader:      reqHeader,
 	})
+	alpaca.CloseResp(ignoredResponse)
```
gnvk, line 32: `ignoredResponse` is actually `nil` here. I'd just keep it as `_` and add a better explanation to the nolint, e.g.: //nolint:bodyclose // According to the websocket.Dial documentation "you never need to close resp.Body yourself"

marketdata/stream/conn_coder.go
```diff
@@ -24,11 +24,12 @@ func newCoderWebsocketConn(ctx context.Context, u url.URL) (conn, error) {
 	reqHeader := http.Header{}
 	reqHeader.Set("Content-Type", "application/msgpack")
 	reqHeader.Set("User-Agent", alpaca.Version())
-	//nolint:bodyclose // According to its docs: you never need to close resp.Body yourself
-	conn, _, err := websocket.Dial(ctxWithTimeout, u.String(), &websocket.DialOptions{
+	//nolint:bodyclose // Linter Error
```
gnvk, line 27: I actually prefer the original here with the clear nolint explanation. In the new version `ignoredResponse.Body` is actually nil, so calling closeResp on it does nothing.

alpaca/rest.go
```diff
@@ -676,7 +661,11 @@ func (c *Client) CancelOrder(orderID string) error {
 		return err
 	}
 
-	return verify(resp)
+	responseVal := Verify(resp)
```
gnvk, line 664: I think it would be slightly clearer if `verify` (just as `unmarshal`) would be responsible for closing the body.

alpaca/rest.go
SahibYar: if I don't make `Verify` exported, I have to copy all the following chunk of code copied to `marketdata` just to keep the logic same. 

```
func verify(resp *http.Response) error {
	if resp.StatusCode >= http.StatusMultipleChoices {
		return APIErrorFromResponse(resp)
	}
	return nil
}
func APIErrorFromResponse(resp *http.Response) error {
	defer closeResp(resp)

	body, err := io.ReadAll(resp.Body)
	if err != nil {
		return err
	}
	var apiErr APIError
	if err := json.Unmarshal(body, &apiErr); err != nil {
		// If the error is not in our JSON format, we simply return the HTTP response
		return fmt.Errorf("%s (HTTP %d)", body, resp.StatusCode)
	}
	apiErr.StatusCode = resp.StatusCode
	apiErr.Body = strings.TrimSpace(string(body))
	return &apiErr
}
func closeResp(resp *http.Response) {
	if resp == nil || resp.Body == nil {
		return
	}
	// The underlying TCP connection cannot be reused if the body is not fully read
	_, _ = io.Copy(io.Discard, resp.Body)
	resp.Body.Close()
}
type APIError struct {
	StatusCode int    `json:"-"`
	Code       int    `json:"code"`
	Message    string `json:"message"`
	Body       string `json:"-"`
}
func (e *APIError) Error() string {
	if e.Code != 0 {
		return fmt.Sprintf("%s (HTTP %d, Code %d)", e.Message, e.StatusCode, e.Code)
	}
	return fmt.Sprintf("%s (HTTP %d)", e.Message, e.StatusCode)
}
```

gnvk: That's fine.

alpaca/rest.go
SahibYar: comment added for explanation.

alpaca/rest.go
SahibYar: actually I am raising another PR, which will remove much of redundant code.

PR #332, opened 2025-03-16
Replace deprecated Dial with DialContext in http.Transport

SahibYar:
## Testing:

Ran `alpaca/stream_test.go` and all tests were passed.


gnvk (approved):
LGTM, thanks!

Issue #292, opened 2024-07-02
Expose rate limit state


gcatlin:
The SDK doesn't expose the http.Response object, so it's tricky to keep track of the rate limit state. Currently I’m logging that info using an `http.RoundTripper`, but it would be good to expose it in a simpler way; e.g., to display the rate limit state in a UI. 

The simplest solution might be to add a callback field to alpaca.ClientOpts that is invoked on every request. It could take a `http.Request` and `http.Response`, or it could take a new object type that contains just the rate limit info from the response. Another option would be to use a `resty.Client` instead of a `http.Client` in the `alpaca.Client`.

My current work-around is to ensure I'm reserving some rate limit buffer when polling the broker API so that I never get a 429 when I need to place orders.

```go
requestsPerTick := 4 // e.g. get account, positions, orders, fills every time we poll
pollingRate := (200.0 - 20.0) / 60.0 // 200 request per min rate limit - 20 requests per min buffer
pollingInterval := time.Duration(float64(time.Second) * float64(requestsPerTick) / pollingRate)
```

Ale-Cas:
@gcatlin is [this](https://github.com/alpacahq/alpaca-trade-api-go/pull/335#issue-2928389327) what you had in mind?

gcatlin:
@Ale-Cas thanks for taking this work on! I'll leave feedback on the PR.

PR #335, opened 2025-03-18
Add user defined callback on http response

Ale-Cas:
resolves https://github.com/alpacahq/alpaca-trade-api-go/issues/292

Screenshot 2025-03-18 at 13 34 47

### DIFF
--- a/alpaca/rest.go
+++ b/alpaca/rest.go
@@ -30,6 +30,8 @@ type ClientOpts struct {
 	RetryDelay   time.Duration
 	// HTTPClient to be used for each http request.
 	HTTPClient *http.Client
+	// ResponseCallback is a user defined callback to execute on the raw http.Response
+	ResponseCallback func(*http.Response)
 }
 
 // Client is the alpaca trading client
@@ -112,6 +114,10 @@ func defaultDo(c *Client, req *http.Request) (*http.Response, error) {
 			break
 		}
 		time.Sleep(c.opts.RetryDelay)
+	}
+
+	if c.opts.ResponseCallback != nil {
+		defer c.opts.ResponseCallback(resp)
 	}
 
 	if err = verify(resp); err != nil {
--- a/marketdata/rest.go
+++ b/marketdata/rest.go
@@ -42,6 +42,8 @@ type ClientOpts struct {
 	HTTPClient *http.Client
 	// Host used to set the http request's host
 	RequestHost string
+	// ResponseCallback is a user defined callback to execute on the raw http.Response
+	ResponseCallback func(*http.Response)
 }
 
 // Client is the alpaca marketdata Client.
@@ -127,6 +129,10 @@ RetryLoop:
 			break
 		}
 		time.Sleep(c.opts.RetryDelay)
+	}
+
+	if c.opts.ResponseCallback != nil {
+		defer c.opts.ResponseCallback(resp)
 	}
 
 	if resp.StatusCode >= http.StatusMultipleChoices {


alpaca/rest.go
```diff
@@ -30,6 +30,8 @@ type ClientOpts struct {
 	RetryDelay   time.Duration
 	// HTTPClient to be used for each http request.
 	HTTPClient *http.Client
+	// ResponseCallback is a user defined callback to execute on the raw http.Response
+	ResponseCallback func(*http.Response)
```
gcatlin, line 34: It would be useful to also provide the request.

```suggestion
	ResponseCallback func(*http.Request, *http.Response)
```

marketdata/rest.go
```diff
@@ -42,6 +42,8 @@ type ClientOpts struct {
 	HTTPClient *http.Client
 	// Host used to set the http request's host
 	RequestHost string
+	// ResponseCallback is a user defined callback to execute on the raw http.Response
+	ResponseCallback func(*http.Response)
```
gcatlin, line 46: Same as above

gcatlin (commented):
I think the callback should also take a `*http.Request` for maximum flexibility. Otherwise, I think this is a reasonable simple solution. Thanks for doing this work!

I don't work at Alpaca, so it's up to them to decide if they want to accept the PR.

alpaca/rest.go
Ale-Cas: Indeed, I agree it can also be useful, I'll wait for the maintainers feedback and implement this accordingly.

gnvk: @gcatlin You're the original reporter of https://github.com/alpacahq/alpaca-trade-api-go/issues/292. If this solution works for you, I'm fine with it with one minor suggestion: if both the request and the response is in the callback then could you maybe call it `HTTPCallback` instead of `ResponseCallback`?

alpaca/rest.go
```diff
@@ -114,6 +116,10 @@ func defaultDo(c *Client, req *http.Request) (*http.Response, error) {
 		time.Sleep(c.opts.RetryDelay)
 	}
 
+	if c.opts.ResponseCallback != nil {
```
gnvk, line 119: The original issue (https://github.com/alpacahq/alpaca-trade-api-go/issues/292) is explicitly about rate limit. However, in this implementation the callback function is not called when the rate limit is hit (L110), only when either we reach the retry limit or succeed. This way there's no way to delay your next request after hitting 429, making the callback unusable for #292.

gcatlin: Thanks for looking at this @gnvk!

The main goal of #292 is to **expose** the rate limit state so we can, for example, make full use of the available request rate for polling (balances, positions, orders, fills) and placing orders, and adjust the polling rate dynamically as needed. This is especially useful when the Alpaca account is accessed from multiple client connections concurrently (e.g. TradingView integration and a locally running trading bot). This PR satisfies that goal.

It looks like it will still call the callback on 429s since L110 only breaks from the for loop but does not return.

alpaca/rest.go
gcatlin: Yeah, this solution works for me. `HTTPCallback` works too.

alpaca/rest.go
gnvk: Even so, I'd call the callback when a 429 happens to cover at least one other use case.

gcatlin: Oh, are you saying we should call the callback the first time we get a 429, rather than waiting until the retry limit is met? If so, I agree.

alpaca/rest.go
Ale-Cas: Will add the request and rename to HTTPCallback than 👍

alpaca/rest.go
Ale-Cas: I changed the implementation and now I'm invoking the method right after receiving the response instead of deferring the execution

gcatlin (commented):
The changes look good. It would also be good to add a test to alpaca/rest_test.go. You could maybe have a flag that gets set when the callback is called and then test against the flag.

PR #339, opened 2025-04-12
Added missing properties to Asset

YannickFuereder:
[Spec](https://docs.alpaca.markets/reference/get-v2-assets-1)

### DIFF
--- a/alpaca/entities.go
+++ b/alpaca/entities.go
@@ -119,16 +119,19 @@ type positionSlice []Position
 type Asset struct {
 	ID                           string      `json:"id"`
 	Class                        AssetClass  `json:"class"`
+	CUSIP                        string      `json:"cusip"`
 	Exchange                     string      `json:"exchange"`
 	Symbol                       string      `json:"symbol"`
 	Name                         string      `json:"name"`
 	Status                       AssetStatus `json:"status"`
 	Tradable                     bool        `json:"tradable"`
 	Marginable                   bool        `json:"marginable"`
-	MaintenanceMarginRequirement uint        `json:"maintenance_margin_requirement"`
 	Shortable                    bool        `json:"shortable"`
 	EasyToBorrow                 bool        `json:"easy_to_borrow"`
 	Fractionable                 bool        `json:"fractionable"`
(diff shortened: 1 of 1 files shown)


alpaca/entities.go
```diff
@@ -119,16 +119,19 @@ type positionSlice []Position
...
 	Symbol                       string      `json:"symbol"`
 	Name                         string      `json:"name"`
 	Status                       AssetStatus `json:"status"`
 	Tradable                     bool        `json:"tradable"`
 	Marginable                   bool        `json:"marginable"`
-	MaintenanceMarginRequirement uint        `json:"maintenance_margin_requirement"`
 	Shortable                    bool        `json:"shortable"`
 	EasyToBorrow                 bool        `json:"easy_to_borrow"`
 	Fractionable                 bool        `json:"fractionable"`
+	MaintenanceMarginRequirement uint        `json:"maintenance_margin_requirement"` // Deprecated: Please use margin_requirement_long or margin_requirement_short instead. Note that these fields are of type string.
```
gnvk, line 132: Let's not add the deprecated field to the SDK!

alpaca/entities.go
```diff
@@ -119,16 +119,19 @@ type positionSlice []Position
...
+	MarginRequirementLong        string      `json:"margin_requirement_long"`
+	MarginRequirementShort       string      `json:"margin_requirement_short"`
```
gnvk, line 134: Both of these should be `decimal.Decimal`.

gnvk (changes requested):
- Please run `make generate` to regenerate the easyjson files.
- Please add / extend a unit test to test the newly added functionality.

alpaca/entities.go
gnvk: Because it's deprecated, which means it should not be used, it only hasn't been removed for the sake of existing users.

gnvk: @YannickFuereder I didn't mean to remove the comment / deprecation warning. I meant to not add the field (`MaintenanceMarginRequirement`).

gnvk: My mistake 🤦 You confused me with changing the order of the fields... In this case what you did was perfect, the deprecation warning should be added (maybe on top of the field instead of next to it).

PR #341, opened 2025-05-05
Add context.Context Support and Improve Context Handling in Stream Methods

molon:
- All `rest` methods now provide proper `context.Context` support
- Clarified behavior of context parameters in streaming methods:
  - In `Connect` method, `ctx` only controls the initial connection process, not the full stream lifecycle
  - To terminate a stream created by `Connect`, use the returned `terminate` function or the `Terminate` method
  - In `StreamTradeUpdatesInBackground`, `ctx` only controls the initial connection, stream continues running independently afterward
  - To stop a stream started by `StreamTradeUpdatesInBackground`, use the returned `terminate` function
- Replaced `time.Sleep` with `ctxtime.Sleep` to ensure timely cancellation on context expiration
- Removed default timeout from `handleSubChange`, now uniformly controlled by `ctx`
- Added buffer to `subChangeRequest.result` to prevent deadlocks caused by `handleSubChange` error returns in edge cases

These changes improve context handling throughout the API, making resource management more predictable and preventing potential deadlocks.


gnvk:
@molon This is a giant backward incompatible change. There are two ways to get this merged:
1. Once reviewed and approved merge it to the v4 branch. There are a couple [more issues labeled with `v4`](https://github.com/alpacahq/alpaca-trade-api-go/issues?q=is%3Aissue%20state%3Aopen%20label%3Av4). Once we do them all, we can consider releasing a new major version of the SDK. One of the benefits of this approach is that we would encourage our users to migrate to the context-aware methods.
2. Instead of adding context to all current client methods, we could duplicate them all. E.g. `client.GetAccount()` to `client.GetAccountCtx(context.Context)`. This way we could keep the major version, however it's an even bigger change. Moreover, it will possibly increase maintenance cost in the future.

molon:
@gnvk 

Sorry for the late reply. I prefer the first way.

Additionally, I'm debating one issue. 
Since I changed the behavior of the ctx parameter in `streaming methods`, even if we go with the first way, should we rename `streaming methods` to make users more aware of this `backward incompatibility`?

zeldahack (approved):

gnvk (changes requested):
This cannot be merged to master (v3) because it contains a lot of breaking changes.

PR #363, opened 2026-03-09
fix: handle partial msgpack map in overnight feed subscription response

gilang-as:
### Summary
  - The overnight feed server returns a msgpack map that declares more fields in the header than it actually encodes on the wire
  - Previously this caused a decode error, preventing successful subscription to the overnight feed
  - Fix by treating io.EOF mid-loop as a graceful end of map instead of an error

### Test plan
  - Subscribe to the overnight feed and confirm no decode error on subscription confirmation
  - Run existing tests: go test ./examples/marketdata/...

### DIFF
--- a/marketdata/stream/message.go
+++ b/marketdata/stream/message.go
@@ -3,6 +3,7 @@ package stream
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"sync"
 
 	"github.com/vmihailenco/msgpack/v5"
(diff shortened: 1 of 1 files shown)


gilang-as:
While trying to use the overnight feed with the sandbox endpoint (DATA_PROXY_WS=wss://stream.data.sandbox.alpaca.markets/v1beta1) using APCA_API_KEY_ID and APCA_API_SECRET_KEY, every subscribe attempt returned an error immediately.

At first I wasn't sure whether I was doing something wrong on my side, or if the overnight feed was not yet supported in the client. After tracing through the code, I found that the subscription confirmation response from the server sends a msgpack map where the header indicates more fields than are actually encoded.

Because of this, DecodeString() returns io.EOF while iterating through the fields, which the client currently treats as a hard error and surfaces as a subscription failure.

In this PR, I handle io.EOF during the map iteration as the end of fields instead of an error. With this change, the client accepts the confirmation response and the subscription works normally.

Please let me know if this behavior is expected or if there is a better way to handle this case.

gnvk:
Thank you for reporting this issue! We fixed the bug on the server side. Please confirm that it's now working (without any client side fix) and close this PR.

gilang-as:
Thanks for the update and for fixing it on the server side. Appreciate the quick fix!

PR #366, opened 2026-05-07
Add US treasuries and corporates endpoints

iupaca:


alpaca/entities.go
```diff
@@ -477,3 +477,165 @@ type optionContractsResponse struct {
...
+	IssuePrice                      decimal.Decimal  `json:"issue_price"`
+	IssueMinimumDenomination        decimal.Decimal  `json:"issue_minimum_denomination"`
+	ParValue                        decimal.Decimal  `json:"par_value"`
+	Callable                        bool             `json:"callable"`
+	NextCallDate                    *civil.Date      `json:"next_call_date,omitempty"`
+	NextCallPrice                   *decimal.Decimal `json:"next_call_price,omitempty"`
+	Puttable                        bool             `json:"puttable"`
+	Convertible                     bool             `json:"convertible"`
+	RegS                            bool             `json:"reg_s"`
+	SPRating                        *string          `json:"sp_rating,omitempty"`
```
gnvk: Somewhat inconsistently other optional strings and enums (e.g. `CouponType` or `Ticker`) are non-pointers, as they should be imo. It makes no sense to differentiate between "" and nil in almost all cases. For the decimals there can be cases where nil and 0 mean different things. 0 dates and times are also questionable, but other entities in the SDK already use pointers for them when they're optional, so let's keep the consistency there.

alpaca/rest_test.go
```diff
@@ -1338,6 +1338,318 @@ func TestGetAccountActivities(t *testing.T) {
...
+	assert.Equal(t, "US912797KJ59", got[0].ISIN)
+	assert.Equal(t, BondStatusOutstanding, got[0].BondStatus)
+	assert.True(t, got[0].Tradable)
+	assert.Equal(t, TreasurySubtypeBill, got[0].Subtype)
+	assert.Equal(t, civil.Date{Year: 2026, Month: 4, Day: 3}, got[0].IssueDate)
+	assert.Equal(t, civil.Date{Year: 2026, Month: 7, Day: 3}, got[0].MaturityDate)
+	assert.Equal(t, "test bill", got[0].Description)
+	assert.Equal(t, "tb", got[0].DescriptionShort)
+	require.NotNil(t, got[0].ClosePrice)
+	assert.True(t, decimal.NewFromFloat(99.6459).Equal(*got[0].ClosePrice))
```
gnvk: * `asset.True` produces undebuggable output
* require.NotNil will fail the whole test
* similar improvements can be made throughout this test and in `TestGetUSCorporates`
```suggestion
	if assert.NotNil(t, got[0].ClosePrice) {
		assert.Equal(t, "99.6459", got[0].ClosePrice.String())
	}
```

gnvk (approved):
LGTM.

PR #368, opened 2026-05-28
Replace float64 with decimal.Decimal in Bar OHLC price fields

carlostasada:
## What

Replace `float64` with `decimal.Decimal` for OHLC price fields in `Bar`, `CryptoBar`, and `OptionBar` market data structs.

## Why

The upstream market data library (`alpaca-trade-api-go`) currently returns `float64` for prices, forcing every downstream consumer to perform `decimal.NewFromFloat()` conversions at the repository boundary. By pushing the decimal type upstream into this library, downstream services receive precise price values directly — eliminating latent IEEE 754 rounding risk in financial calculations.

This is a **companion PR** — the main work in the portfolio history service repo depends on this library change.

## Changes

- `Bar.Open`, `Bar.High`, `Bar.Low`, `Bar.Close` → `decimal.Decimal` (was `float64`)
- `CryptoBar.Open`, `CryptoBar.High`, `CryptoBar.Low`, `CryptoBar.Close` → `decimal.Decimal` (was `float64`)
- `OptionBar.Open`, `OptionBar.High`, `OptionBar.Low`, `OptionBar.Close` → `decimal.Decimal` (was `float64`)
- `VWAP` remains `float64` — it is a computed weighted average, not a discrete price quote
- `CryptoBar.Volume` remains `float64` — it is a quantity, not a monetary value
- Regenerated `entities_easyjson.go` to match updated struct field types

## Testing

- Struct type changes are compile-time enforced — any consumer using the old `float64` type will fail to build, making regressions impossible to miss
- `easyjson` serialization regenerated and verified for correct `decimal.Decimal` JSON marshaling (shopspring/decimal implements `json.Marshaler`/`json.Unmarshaler`)
- No behavioral tests in this repo for these structs — correctness validated in the downstream service tests

## Desired feedback

- **Breaking change scope**: This is intentionally breaking for all consumers of `Bar`, `CryptoBar`, and `OptionBar` OHLC fields. Please verify the scope is acceptable and that we've communicated the break to affected teams.
- **VWAP decision**: Left as `float64` since it's a computed average rather than a quoted price. Does this distinction make sense, or should VWAP also be `decimal.Decimal` for consistency?
- **Volume field**: `CryptoBar.Volume` stays `float64` (fractional crypto quantities). Correct call?

## Reviewer guide

Suggested read order:
1. Start with: `marketdata/entities.go` — the actual type changes (3 structs)
2. Skip: `marketdata/entities_easyjson.go` — auto-generated, do not review line-by-line

## Checklist (author self-review)

- [x] Read the diff top-to-bottom as a reviewer would
- [x] No debug code, commented-out blocks, or TODO items left in
- [x] Secrets / credentials not committed
- [x] If financial path: decimal types used, no float arithmetic
- [x] If API contract changes: consumers notified or backward-compatible
- [ ] Companion PR in portfolio history service repo links this change

---

## Reviewer Briefing

**TL;DR**: OHLC price fields in 3 bar structs changed from `float64` to `decimal.Decimal`. Breaking change for all consumers. Verify the boundary decision (VWAP/Volume excluded) is correct.

### Review order (highest to lowest priority)

| Priority | File | Risk | Why |
|---|---|---|---|
| 1 | `marketdata/entities.go` | HIGH | Financial type change — `Bar`, `CryptoBar`, `OptionBar` OHLC fields now `decimal.Decimal`. Breaking public API. |
| 2 | `marketdata/entities_easyjson.go` | LOW | Auto-generated serialization code — do not review line-by-line |

### Danger zones

- **`marketdata/entities.go` (lines 181–196, 239–254, 539–554)**: The three bar structs. Verify that ONLY price fields (Open/High/Low/Close) are changed and that VWAP/Volume remain untouched. A mistaken type change on Volume would break quantity semantics.

### Safe to approve quickly

- `marketdata/entities_easyjson.go` — generated file, matches struct changes mechanically

### Open questions for reviewer

- Should `VWAP` also become `decimal.Decimal` for API consistency, even though it's a computed weighted average?
- Are there other consumers of this library beyond the portfolio history service that need advance notice of this breaking change?
- Does `marketcache` (the other upstream source mentioned in the ticket) also need a parallel PR?

---


gnvk:
@carlostasada This would be a breaking change. We can only even consider it for the next major version (`v4`). But honestly I don't think it's necessary at all. Actual issues with float64s can only arise with many fractional digits or very large integers, both of which are practically impossible for real bar prices.

PR #370, opened 2026-07-16
Remove US treasuries and corporates endpoints

carlostasada:
The `assets/fixed_income/us_treasuries` and
`assets/fixed_income/us_corporates` endpoints have been removed from the Trading API, so the corresponding SDK methods can no longer function.

Remove the now-dead code:
- `GetUSTreasuries` / `GetUSCorporates` client methods and package-level wrappers, plus their request structs
- `USTreasury` / `USCorporate` entities, their response structs, and the enums used only by them (`BondStatus`, `TreasurySubtype`, `CouponType`, `CouponFrequency`, `DayCount`, `SPOutlook`, `CallType`)
- the associated tests, and regenerate `entities_easyjson.go`

BREAKING CHANGE: `GetUSTreasuries`, `GetUSCorporates`, and the `USTreasury`/`USCorporate` types (and their supporting enums) are removed. Callers of these APIs must remove their usages.


said-saifi (approved):

gnvk (commented):
I disagree with this change, but won't block it.

carlostasada:
Those endpoints are still valid, only temporarily hidden and probably added back in the future

PR #371, opened 2026-07-22
Add raw message handler to the market data stream

LQ-Epoch:
Adds an opt-in callback invoked once for every exact MessagePack WebSocket frame before the bounded processing buffer. The callback receives an owned copy, enabling audit envelopes and protocol-drift detection without replacing the official decoder. Includes an offline ownership/order test.\n\nThis is intentionally a minimal hook; existing behavior is unchanged when unset.


marketdata/stream/client.go
```diff
@@ -606,6 +608,9 @@ func (c *client) connReader(
 			}
 			return
 		}
+		if c.rawMessageHandler != nil {
+			c.rawMessageHandler(append([]byte(nil), msg...))
```
gnvk, line 612: [minor] `slices.Clone(msg` or `bytes.Clone(msg)` is functionally the same but more readable.

marketdata/stream/client.go
```diff
@@ -606,6 +608,9 @@ func (c *client) connReader(
...
+		}
```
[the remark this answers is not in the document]
gnvk: I think this is acceptable.

marketdata/stream/options.go
```diff
@@ -135,6 +136,20 @@ func WithBufferFillCallback(callback func(msg []byte)) Option {
 	})
 }
 
+// WithRawMessageHandler runs handler once for every complete WebSocket frame,
+// before the frame enters the client's processing buffer. The byte slice is an
+// exact copy of the received MessagePack frame and is owned by the handler.
```
gnvk, line 141: We might want to consider giving the user the option to get the original byte slice instead of copying it for every single frame. If the handler doesn't modify it then the extra copies are pointless overhead. Maybe
``` go
WithRawMessageHandler(handler func(msg []byte), passOriginalMessage bool)
```

marketdata/stream/raw_message_test.go
```diff
@@ -0,0 +1,41 @@
...
+	workers.Add(1)
+	closeCh := make(chan struct{})
+	go client.connReader(ctx, &workers, closeCh)
+
+	want := []byte{0x91, 0x81, 0xa1, 'T', 0xa1, 'q'}
+	connection.readCh <- want
+	processed := <-client.in
+	got := <-captured
+	processed[0] = 0
+	if !bytes.Equal(got, []byte{0x91, 0x81, 0xa1, 'T', 0xa1, 'q'}) {
```
gnvk, line 30: [minor] Please use `assert.Equal`

marketdata/stream/raw_message_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"bytes"
+	"context"
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestRawMessageHandlerReceivesOwnedExactFrameBeforeBuffer(t *testing.T) {
+	connection := newMockConn()
+	client := newClient()
```
gnvk, line 13: [minor] Please use the exported API to create the client
``` go
client := NewStocksClient(
	marketdata.IEX,
	WithRawMessageHandler(func(frame []byte) {
		captured <- got
	}),
	withConnCreator(func(_ context.Context, _ url.URL) (conn, error) {
		return connection, nil
	}),
)
```

marketdata/stream/raw_message_test.go
```diff
@@ -0,0 +1,41 @@
+package stream
```
gnvk, line 1: [minor] I don't think it makes sense to create a dedicated file for just this one test case, let's put it to client_test.go.

marketdata/stream/raw_message_test.go
```diff
@@ -0,0 +1,41 @@
...
+		t.Fatalf("raw callback frame changed: %x", got)
+	}
+
+	cancel()
+	select {
+	case <-closeCh:
+	case <-time.After(time.Second):
+		t.Fatal("reader did not stop")
+	}
+	workers.Wait()
```
gnvk, line 40: [minor] I don't think these are relevant for the test, they can be skipped

PR #376, opened 2026-09-10
CI: pin Actions to SHAs and add Dependabot for Actions and Go modules

carlostasada:
Pinning CI to immutable commit SHAs prevents a moved or compromised tag from changing what runs. Dependabot keeps those SHAs and Go modules current in weekly grouped PRs.


gnvk (approved):

.github/workflows/go.yml
```diff
@@ -13,20 +13,20 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5.6.0
         with:
           go-version: 1.23
-      - uses: golangci/golangci-lint-action@v6
+      - uses: golangci/golangci-lint-action@55c2c1448f86e01eaae002a5a3a9624417608d84 # v6.5.2
```
gnvk, line 29: golangci-lint should be upgraded to v2, but it's a sizable change of course.

carlostasada: I will deal with it in a future PR

